PR #51, opened 2015-03-13
refactor(symbol): renamed Symbol to AMQPSymbol

mbroadst:
Symbol is a reserved word in [ES6](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol), so while this is not a problem today
you will run into issues in the future. Incidentally, jshint and the like
are by default checking for this so this will help while using those
tools.

I've also included a Makefile here as a lightweight replacement for your Grunt setup. If you're amenable, I can update the Makefile to cover all your Grunt tasks as a secondary PR (I also have one incoming after this for a generalized cleanup using the closure linter)


mbroadst:
Ah, looks like grunt is even screwing up your travis builds :smile:

PR #63, opened 2015-03-14
refactor(buffers): use buffer lists instead of a circular buffer

mbroadst:
Closes https://github.com/noodlefrenzy/node-amqp-1-0/issues/35, possibly https://github.com/noodlefrenzy/node-amqp-1-0/issues/47 as well

This was a pretty straight-forward port, and seems to have resolved the parsing issues I was seeing previously. I've intentionally left the sort of awkward combination of "slice(0, offset); consume(offset)" in the code as a sort of reminder that we should eventually be using streams/token-parser, but this is a good stopgap for the moment.

### DIFF
--- a/lib/codec.js
+++ b/lib/codec.js
@@ -3,7 +3,6 @@
 var debug = require('debug')('amqp10-Codec'),
     util = require('util'),
     butils = require('butils'),
-    CBuffer = require('cbarrick-circular-buffer'),
     builder = require('buffer-builder'),
     Int64 = require('node-int64'),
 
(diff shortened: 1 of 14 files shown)


mbroadst:
@noodlefrenzy okay we should be all good to go, might I also suggest you bump the version? (I need to use this fix in my project)

noodlefrenzy:
Pushed 0.0.15 to npm with these changes.

noodlefrenzy:
Also, re: your name changes - those are great, I like them.  Re: the debug output, I'm fine keeping it in (or, now, putting it back) for a while and ripping it out when we do a wider sweep/cleanup.

PR #67, opened 2015-03-17
show how lodash can be used to replace deepMerge

mbroadst:
deepMerge was used for much more than it claimed to do, so this
commit is meant to show how lodash can replace it. It's not meant
to be merged, but rather as a cheat sheet for how uses of deepMerge
can be replaced in the codebase


mbroadst:
Remember, [merge-defaults](https://github.com/balderdashy/merge-defaults) can help with deep _.defaults support. Merging right is indeed not supported in lodash, but I might argue for convention in this case

noodlefrenzy:
Yep yep, just letting you know, because I was expecting to finish it last night and now it's going to take slightly longer.  Also, defaults is destructive, and deepMerge wasn't, so need to determine when that'll be a problem. 

As for merge-right convention - I'll obviously switch all of my internal invocations to match lodash's convention, but I've exposed merge-right as an API method on AMQPClient.policies, so I'll leave that as-is.

noodlefrenzy:
Oh, and _.merge is super destructive - it destroys all but the primary source.

mbroadst:
@noodlefrenzy ah yeah I was just throwing those in to get the tests passing. It might be a useful refactor to rename deepMerge into a few different methods (deepClone, deepMerge, etc), to better understand what the intent is in each place its used before moving to the lodash methods? Sounds like most of those cases  you were using an equivalent of _.clone and _.cloneDeep?

noodlefrenzy:
@mbroadst Yep, that's exactly my plan (re: refactor to appropriately named methods that delegate to lodash appropriately).

PR #68, opened 2015-03-18
refactor(frame_reader): reduce complixity and enhance readability

mbroadst:
FrameReader used to be a large collection of switch statements. This
change reduces that complexity by splitting into the concept of AMQP
and SASL frame internal reader classes. A number of other readability
cleanups occurred including: unrolling a number of if statements to
fail early, as well as registration of Readers.

**NOTE:** I'd like to further discuss this PR, because while the readability of having multiple internal FrameReaders is nice, this can be optimized even more to create a single registry in the FrameReader class encompassing all frame types we accept. It would reduce readability imho, but probably improve performance. Furthermore, if we can agree on what all *_frame.js files should look like, we can automate the building of the frame registry.

### DIFF
--- a/lib/frames/frame_reader.js
+++ b/lib/frames/frame_reader.js
@@ -2,6 +2,9 @@
 
 var debug = require('debug')('amqp10-FrameReader'),
 
+    _ = require('lodash'),
+    util = require('util'),
+
     Attach = require('./attach_frame'),
     Begin = require('./begin_frame'),
     Close = require('./close_frame'),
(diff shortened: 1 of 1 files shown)


mbroadst:
alright, how about that?

noodlefrenzy:
As for your note, I'm unwilling to sacrifice the readability of distinct frame classes for perf... yet.  However, I definitely realize that there is quite a bit of cleanup that could happen there.  Perhaps the distinct classes could, instead of providing the readPerformative, provide a field array and a defaults object, we could build the reader from that.

mbroadst:
@noodlefrenzy well I was more speaking about the possibility of doing a further refactoring to merge AmqpFrameReader and SaslFrameReaders into a single lookup object, a la:

```
var registry = {
  '0x00': {       // AMQP Frame
     'open-descriptor-whatever-it-is': Open,
     'close-descriptor...': Close
  },

  '0x01': {      // SASL Frame
     // you get the idea
  }
}

```

mbroadst:
@noodlefrenzy either way, it's sort of a problem for the future, I ran some benchmarks last night and it didn't buy me _that_ much

noodlefrenzy:
Ah, that, yeah I'm totally cool with that :)

noodlefrenzy:
Want me to cut a new release with these changes?

mbroadst:
I'm writing a slew of unit tests to get full coverage for the frame reader, I'd say new release after that?

noodlefrenzy:
Fantastic!  Sounds good to me.

PR #72, opened 2015-03-23
chore(project-name): rename project to node-amqp10

mbroadst:
As discussed this change introduces a project name change. The
proposed benefits are as follows:

  - name reflects node best practices
  - easier to find the project on npmjs and google

It is early enough in the projects life that this should not cause
too much disruption, so now seems a better time than later to make
the change.

**NOTE:** This will require a little more work in terms of actually renaming the project on github, travis and code climate. I also suggest a bump in version (perhaps a minor version bump, or even *gasp* the dreaded major version bump :smile:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-node-amqp-1-0
+node-amqp10
 =============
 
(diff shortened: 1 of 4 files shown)


mbroadst:
@noodlefrenzy okay I think thats all of them

PR #73, opened 2015-03-23
refactor(test-send-buffers): streamline outgoing buffers for tests

mbroadst:
A few things going on here:

 - remove all the temporary <frameType>buf() methods from tests,
   leading to a more declarative form for what the expected
   requests and responses are for a test sequence
 - removal of the Frame.outgoing helper function

This is the first step towards removing the buffer-builder from
the codebase. Since we are working with streams, there should be
no need to actually use a buffer builder. Now that this is in
place, we are able to use modules like buffer-more-ints to work
directly with streams (initially) for outgoing encoding.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -329,10 +329,8 @@ Connection.prototype.close = function() {
 };
 
 Connection.prototype.sendFrame = function(frame) {
-  var frameBuf = frame.outgoing();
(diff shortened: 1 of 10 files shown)


mbroadst:
I know I'm sort of growing the dependency tree here, but such is the price of slow refactoring :smile:

PR #77, opened 2015-03-25
test(encoding-decoding): improve test coverage for encoding/decoding

mbroadst:
First shot at complete coverage for all currently supported scalar
and collection types. Switched to using ForcedType's for all the
encoding to make sure we're covering what we think we're covering.
There were a few changes made to tests, mostly where null values
were being encoded incorrectly as full values rather than their
<type>-null variant.


mbroadst:
@noodlefrenzy updated to include some commented out tests, and to add missing tests to verify e.g. that Int64s are accepted (per discussion)

mbroadst:
@noodlefrenzy not sure what was up with that, but the 0x44 fix in as well now

PR #85, opened 2015-05-21
fix(message-properties): add missing field and correct null values

mbroadst:
Message.Properties omitted the "correlation-id" field, and did not
properly handle undefined values for the following fields:
  - absolute-expiry-time
  - creation-time
  - group-sequence

### DIFF
--- a/lib/types/message.js
+++ b/lib/types/message.js
@@ -180,15 +180,19 @@ Properties.prototype.getValue = function() {
     to: u.orNull(self.to),
     subject: u.orNull(self.subject),
     replyTo: u.orNull(self.replyTo),
+    correlationId: u.orNull(self.correlationId),
     contentType: u.orNull(self.contentType),
     contentEncoding: u.orNull(self.contentEncoding),
-    absoluteExpiryTime: new ForcedType('timestamp', self.absoluteExpiryTime),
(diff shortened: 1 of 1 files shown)


mbroadst:
I'd also like to discuss a proposed change for described types. Currently there is tons of duplication in this code based around a shared "schema" (the name I'm using for the moment). If we take "Properties" as an example, I would propose that we add something like this to the definition:

```
Properties.Schema = {
  messageId: 'variant';           /* ulong, uuid, binary, string */
  userId: 'binary';
  to: 'string';
  subject: 'string';
  replyTo: 'string';
  correlationId: 'variant';      /* ulong, uuid, binary, string */
  contentType: 'symbol';
  contentEncoding: 'symbol';
  absoluteExpiryTime: 'timestamp';
  creationTime: 'timestamp';
  groupId: 'string';
  groupSequence: 'uint';
  replyToGroupId: 'string';
};
```

The benefit of having a schema is that we can create very easily shared code for fromDescribedType, getValue, as well as the constructor. As an added benefit, we could easily automate the inclusion/exclusion of values in the encodeOrdering (currently we just send EVERY value, even if its null, and this is not required as far as my investigations have gone).

Thoughts? Also, I'm not quite sure what to do about the "variant" values in the schema above.

PR #90, opened 2015-05-22
refactor(connect): remove side-effecty connect calls

mbroadst:
There were a number of locations where connect was called in the
event that a user had not previously connected. As this is prone
to error, it's being removed in favor forcing users to call connect
explicitly.

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -64,16 +64,9 @@ var EventEmitter = require('events').EventEmitter,
  * instantiation to sending messages.
  *
  * @param {PolicyBase} [policy]     Policy to use for connection, sessions, links, etc.  Defaults to PolicyBase.
- * @param {string} [uri]            If provided, must provide cb.  Will attempt connection, set default queue.
- * @param {function} [cb]           If provided, must provide uri.  Will attempt connection and call cb when established/failed.
  * @constructor
  */
-function AMQPClient(policy, uri, cb) {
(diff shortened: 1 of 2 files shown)


noodlefrenzy:
Can remove:
'''
  // If given full address, pull out target first to avoid leaking credentials in error messages containing linkName.
  var rootUri;
  if (target && target.toLowerCase().lastIndexOf('amqp', 0) === 0) {
    var address = u.parseAddress(target);
    rootUri = address.rootUri;
    target = address.path.substring(1);
  }
'''
from send/receive, since we're no longer allowing them to send/receive without calling connect first.

noodlefrenzy:
Also, just to clarify - the connection code in send/receive is there for four reasons:
1) So users can call send/receive without calling connect
2) So send/receive can automagically reconnect when disconnected
3) So a user can call receive on 30 queues all at once without waiting for links to be established one at a time
4) So a user can call send with 50 messages all at once without waiting for an initial link to be established

1 and 2 can be dropped, I'm fine with that as long as the client has some way to know that they need to reconnect (especially in the case of receivers).
We need to make sure 3 and 4 still work. 3 should work fine, I believe, since receivers are all on different links and just need to make sure that the connection and session is established (which is completed once connect returns). 4 is the harder one, since all sends go over the same sender link. Looking at the refactored code I'm fairly confident this still works, and I believe I have a test for this case, so I'm not that worried. Mostly just documenting for posterity.

PR #92, opened 2015-05-23
refactor(send): promisify send

mbroadst:


noodlefrenzy:
Clean up / respond to the two (!(!!err)) comments above, but other than that looks good. Really like the promisified syntax so much better that cb-soup.

PR #114, opened 2015-06-18
policy refactor

mbroadst:
* remove "Policy" suffixes, as they are redundant
* rename PolicyBase to DefaultPolicy
* export all policies on the top level of the module as e.g. ```Policy.Default```, ```Policy.EventHub```


noodlefrenzy:
... or I could just read the rest of the review.

PR #128, opened 2015-07-22
feat(unique-link-names): use uniquely generated link names

mbroadst:
This follows more closely to what QPID is doing with their messaging framework, in that link names by default are generated based on their address (and in our case link creation properties). A hash of the address and options is used to uniquely identify an attempt to recreate an existing receiver.

The user is still able to override this behavior by passing in a "name" parameter to the link creation "options" object (verified in the unit tests)


lib/amqp_client.js
```diff
@@ -506,13 +534,14 @@ AMQPClient.prototype.createReceiver = function(address, options, cb) {
     }
   }
 
-  var linkName = address + '_RX';
-  if (this._attaching[linkName]) {
-    self._onReceipt[linkName].push(cb);
+  var linkInfo = self._linkInfo('receiver', address, options);
+
+  if (this._attaching[linkInfo.hash]) {
```
noodlefrenzy: Might as well s/this/self/

mbroadst: updated

lib/amqp_client.js
```diff
@@ -81,6 +81,11 @@ function AMQPClient(policy) {
   this._pendingSends = {};
   this._unsettledSends = {};
 
+  this._linkNameForHash = {
```
noodlefrenzy: I'd change this to linkNamesByHash...

lib/amqp_client.js
```diff
@@ -219,14 +224,36 @@ AMQPClient.prototype.connect = function(url) {
...
+  var name = u.linkName(address, options);
+  lookup[hash] = name;
+  return { name: name, hash: hash };
+};
+
+AMQPClient.prototype._resolveDeferredSenderAttaches = function(linkHash, err, link) {
   // resolve deferred attachments
-  while (this._senderAttaching[linkName].length) {
-    var deferredAttach = this._senderAttaching[linkName].shift();
+  while (this._senderAttaching[linkHash].length) {
```
noodlefrenzy: ... and then for any map indexed by hash add ByHash to the name

Otherwise, this is just a bug-farm.

lib/amqp_client.js
```diff
@@ -219,14 +224,36 @@ AMQPClient.prototype.connect = function(url) {
...
+ * @param address   the address of the link
+ * @param options   the link creation options
+ * @returns {Object}
+ */
+AMQPClient.prototype._linkInfo = function(type, address, options) {
+  var lookup = (type === 'receiver') ?
+    this._linkNameForHash.receivers : this._linkNameForHash.senders;
+
+  var hash = u.linkHash(address, options);
+  if (lookup.hasOwnProperty(hash)) {
```
noodlefrenzy: It's now possible for a client to (try and) create two links with the same name if they pass in the same options.name for differently-hashed values. I don't think that was possible before.

Adding a this._linkNamesByHash.reverseMap[name] = hash;, checking it, and throwing an error on that attempt wouldn't be hard to do, and you wouldn't need/want separate sender/receiver maps so not much extra overhead.

PR #129, opened 2015-07-23
Remove send from amqpclient

mbroadst:
Continuing from the initial commit, this fully removes "send" from the Client api and modifies the tests accordingly. Additionally, the mock classes in the AMQPClient test have been refactored out into their own modules.


test/unit/mocks/sender_link.js
```diff
@@ -0,0 +1,41 @@
...
+  this.session = session;
+  this.options = options;
+  this._clearState();
+}
+
+util.inherits(MockSenderLink, SenderLink);
+
+MockSenderLink.prototype.canSend = function() {
+  this.emit('canSend-called', this);
+  return this.capacity > 0;
```
noodlefrenzy: I see what you mean re: capacity. I was only using this because it was easier (for testing) than tracking session/link credits - we might be able to get away with using those now and just calling base.canSend here.

test/unit/mocks/sender_link.js
```diff
@@ -0,0 +1,41 @@
...
+  var self = this;
+  self.curId++;
+  self.messages.push({ id: self.curId, message: msg.body, options: options });
+  self.emit('sendMessage-called', self, self.curId, msg, options);
+  return self.curId;
+};
+
+MockSenderLink.prototype._clearState = function() {
+  this.name = this.options.name;
+  this.isSender = this.options.isSender || false;
```
noodlefrenzy: Isn't this trivially true now that you've split sender/receiver link classes?

test/unit/test_amqpclient.js
```diff
@@ -153,136 +50,68 @@ describe('AMQPClient', function() {
...
+        .then(function () {
+          return client.createSender(queue);
         })
-        .then(function() {
+        .then(function (sender) {
+          return Promise.all([
+            sender.send({ my: 'message' }),
+            process.nextTick(function() {
+              sender.capacity = 100;
+              sender._dispatchPendingSends();
```
noodlefrenzy: Yeah, this kinda sucks vs. the flowframe way that actually triggers the code.

test/unit/test_amqpclient.js
```diff
@@ -685,16 +210,16 @@ describe('AMQPClient', function() {
 
 
     it('should create connection, session, and link on receive with full address', function() {
```
noodlefrenzy: This is no longer true. Should consider either renaming, or just drop the test if it's covered by other things.

test/unit/test_amqpclient.js
```diff
@@ -732,15 +257,15 @@ describe('AMQPClient', function() {
     });
 
     it('should only create a single connection, session, multiple links for multiple receives', function() {
```
noodlefrenzy: This test also doesn't really make any sense anymore since creating a link no longer spawns the connection/session

test/unit/mocks/sender_link.js
mbroadst: well the base canSend still uses remote numbers.. I think for now capacity is okay, as long as we agree that moving forward we'll try to use as much of the base logic as possible

test/unit/mocks/sender_link.js
mbroadst: absolutely, I'll drop that

test/unit/test_amqpclient.js
mbroadst: chalking this up to "I'm about to make a sender_link.test.js, so let's worry about it there"

test/unit/test_amqpclient.js
mbroadst: I'm going to break apart this class, so I'll remove/reword these tests at that point, unless you think its important for this particular PR

PR #131, opened 2015-07-24
Servicebus integration test - just getting started.

noodlefrenzy:
Encoded creds in .travis.yml, initial send/receive test for ServiceBus queues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,39 +1,28 @@
 language: node_js
-
 node_js:
-  - "0.10"
-  - "0.12"
-
+- '0.10'
+- '0.12'
 cache:
   directories:
-    - node_modules
-
+  - node_modules
 branches:
   except:
-    - gh-pages
-
-# addons:
-#   apt:
-#     sources:
-#       - rabbitmq-testing
-#       - qpid-testing
-#     packages:
-#       - rabbitmq-server
-#       - qpidd
-
+  - gh-pages
 before_script:
-  - npm install -g codeclimate-test-reporter
-  - sudo add-apt-repository ppa:mcpierce/qpid-testing -y
-  - sudo apt-get update -q
-  - sudo apt-get install qpidd qpid-tools
-  - sudo sh -c 'echo "auth=no" >> /etc/qpid/qpidd.conf'
-  - sudo /etc/init.d/qpidd restart
-
-  # remove these when we can specify link properties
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,39 +1,28 @@
 language: node_js
-
 node_js:
-  - "0.10"
-  - "0.12"
-
+- '0.10'
```
mbroadst: a lot of these indentations seem incorrect

lib/types/message.js
```diff
@@ -379,7 +379,13 @@ Message.prototype.encode = function(codec, buf) {
   if (this.properties) codec.encode(this.properties, buf);
   if (this.applicationProperties) codec.encode(this.applicationProperties, buf);
   if (this.body.length > 1) {
-    codec.encode(new AMQPSequence(this.body), buf);
+    if (this.body instanceof Buffer) {
```
mbroadst: whoops, my bad :)

test/integration/servicebus/config.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+module.exports = {
+
```
mbroadst: we should investigate a way to unify the integration test config files, so like protocol, host, address/target, username, password are all sort of generic? (I'm saying just a future thought, not for this PR).

The test-fixture class could eventually provide an easy means for combining these into the "config.address" variable, etc.

mbroadst:
only one actual comment, otherwise LGTM

.travis.yml
noodlefrenzy: Fixed. I was using travis commands to add the secure keys, and it mucked with the formatting.

test/integration/servicebus/config.js
noodlefrenzy: I agree, ish. I think address makes sense, and defaultLink/queueName/whatever makes sense, but beyond that I'm less sure. I'm hoping to write some tests for topics this week - those are different enough to muck up any assumptions about commonalities, I'm hoping.

.travis.yml
```diff
@@ -12,15 +11,6 @@ branches:
   except:
     - gh-pages
 
-# addons:
```
mbroadst: not to be too nitpicky, but these are useful comments for me considering Travis is currently in the final stages of approving my request to whitelist the qpid-testing apt-source

PR #133, opened 2015-07-27
feat(default-subject): allow specifying subjects in link names

mbroadst:
This patch adds support for defining a subject in the name of a
link during link creation. On the receiver side this adds a filter
for the subject, and on the sender side this adds a default subject
to the message properties (if one does not currently exist).


lib/amqp_client.js
```diff
@@ -378,11 +379,23 @@ AMQPClient.prototype.createReceiver = function(address, options, cb) {
...
+      source: { address: address.name },
       target: { address: 'localhost' }
     }
   }, self.policy.receiverLink);
 
+  // if a subject has been provided then automatically set up a filter to match
+  // on that subject.
+  if (!!address.subject) {
+    var filterSymbol = (address.subject.indexOf('*') || address.subject.indexOf('#')) ?
+      'apache.org:legacy-amqp-topic-binding:string' :
```
noodlefrenzy: I question how many servers actually adhere to this format, given that it's not part of the spec.

noodlefrenzy: I guess my point is that this seems like it might be something that could be factored into a functor inside a qpid policy, with functors for the other policies that "did the right thing" with their "subjects"

mbroadst: http://www.amqp.org/specification/1.0/filters

mbroadst: or do you mean how the actual value is put together?

noodlefrenzy: Ah, ok, if it's registered as a non-normative extension then that's fine to encode it in there. Perhaps a comment with the link to that, though.

mbroadst: I think I'll move them into an exported module, since I'd like to be able to reuse these values for the short-hand filter format we discussed earlier today. Perhaps `require('amqp10').Filters` ?

lib/utilities.js
```diff
@@ -145,6 +145,14 @@ function parseAddress(address) {
 
 module.exports.parseAddress = parseAddress;
 
+// @todo: this "parsing" should be far more rigorous
+module.exports.parseLinkAddress = function(address) {
+  var parts = address.split('/');
```
noodlefrenzy: Actually, the defaultSubjects policy might come into play here instead (i.e. don't attempt to segment if we don't cope with subjects). I'll know more once I write the EH integration test.

mbroadst: good point, I'll throw it in a level up

mbroadst:
closed in favor of #143

PR #134, opened 2015-07-28
fix(disposition): only send disposition to relevant links by role

mbroadst:
Previously we ignored the role of the incoming disposition frame,
and forwarded these frames to the _dispositionRecieved method of
all links we knew about. This patch corrects this behavior, and
includes an integration test verifying that frames are forwarded
to the appropriate link types.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -119,8 +119,10 @@ function Session(conn) {
   this._allocatedHandles = {};
   this._linksByName = {};
   this._linksByRemoteHandle = {};
-  this._senderLinks = [];
   this._deliveryTag = 1;
+
+  this._senderLinks = [];
+  this._receiverLinks = [];
 
   var self = this;
   var stateMachine = {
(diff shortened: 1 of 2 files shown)


lib/session.js
```diff
@@ -414,9 +417,12 @@ Session.prototype._handleDisposition = function(frame) {
     state: frame.state
   };
 
-  _.values(this._linksByName).forEach(function(link) {
-    link._dispositionReceived(disposition);
-  });
+  var dispositionHandler = function(l) { l._dispositionReceived(disposition); };
+  if (frame.role === constants.linkRole.sender) {
+    this._senderLinks.forEach(dispositionHandler);
```
noodlefrenzy: I think this is actually reversed - the "role" is the role of the one sending you the dispo frame (so e.g. "I'm a receiver, and I'm sending this dispo frame to the sender").

mbroadst: oh, indeed you are correct. The spec is very mysterious in their wording here, but it should have been obvious from the test that the sender should be the one receiving dispositions!

PR #136, opened 2015-07-31
feat(Client): remove receipt callback on createReceiver

mbroadst:
This greatly simplifies the internals for receiver link creation,
aligning it more closely with sender link creation (so we can begin
to share code between them). It also makes the API more explicit by
requiring the user to understand that the receiver link is indeed
an EventEmitter and should be used that way.

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -77,7 +77,6 @@ function AMQPClient(policy) {
   this._attaching = {};
   this._senderAttaching = {};
   this._attached = {};
-  this._onReceipt = {};
   this._pendingSends = {};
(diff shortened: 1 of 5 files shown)


lib/amqp_client.js
```diff
@@ -391,7 +374,7 @@ AMQPClient.prototype.createReceiver = function(address, options, cb) {
     }
 
     if (options.policy) {
-      u.deepMerge(linkPolicy, options.policy);
+      linkPolicy = u.deepMerge(linkPolicy, options.policy);
```
noodlefrenzy: holy crap, good catch

PR #137, opened 2015-08-01
Link creation cleanup

mbroadst:
The general idea here was to expose how similar createSender and createReceiver actually were. By cleaning up, and refactoring these methods, it's pretty clear that they share most of the same code. Additionally, I opted to listen directly to created links for events like `Link.Attached` in order to get around issues we had previously with needing to do string checks for incoming attached events.

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -205,6 +205,16 @@ AMQPClient.prototype._resolveDeferredSenderAttaches = function(linkName, err, li
   this._senderAttaching[linkName] = undefined;
 };
 
+AMQPClient.prototype._resolveDeferredReceiverAttaches = function(linkName, err, link) {
+  // resolve deferred attachments
+  while (this._attaching[linkName].length) {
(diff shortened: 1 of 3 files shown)


lib/amqp_client.js
```diff
@@ -249,47 +266,33 @@ AMQPClient.prototype.createSender = function(address, options) {
...
-          options: {
-            name: linkName,
-            source: { address: 'localhost' },
-            target: { address: address }
-          }
-        }, self.policy.senderLink);
-        self._session.attachLink(linkPolicy);
+
+        var link = self._session.attachLink(linkPolicy);
+        link.on(Link.Attached, onAttached);
```
noodlefrenzy: Man, why didn't I write it this way originally?

mbroadst: I don't necessarily disagree with the original design choice, a lot of the cleanup I'm doing ends up in increased coupling. We should talk about it. But yeah, this definitely solves the outstanding issue for the uuid link names :smile:

PR #140, opened 2015-08-03
refactor(links): move responsibility of Links fully to Session

mbroadst:
There are some cleanups here (removing events that shouldn't be used anymore, generally going back on my stupid decision to put namespaces in event names, etc), but more importantly is the change to remove attaching/attached tracking for links inside AMQPClient. It is now the responsibility of a Session to track Link state, and the Client's create[Sender,Receiver] methods are merely proxies at this point to Session.createLink.

### DIFF
--- a/lib/link.js
+++ b/lib/link.js
@@ -22,6 +22,8 @@ function Link(session, handle, linkPolicy) {
   this.attached = false;
   this.remoteHandle = undefined;
 
+  this._onAttach = [];
+
   var self = this;
   var stateMachine = {
     'DETACHED': {
@@ -69,14 +71,14 @@ Link.MessageReceived = 'message';
 Link.ErrorReceived = 'error';
 
 // On link credit changed.
-Link.CreditChange = 'link:creditChange';
+Link.CreditChange = 'creditChange';
 
 
 // On completion of detach.
-Link.Attached = 'link:attached';
+Link.Attached = 'attached';
 
 // On completion of detach.
-Link.Detached = 'link:detached';
+Link.Detached = 'detached';
 
 // public api
 Link.prototype.attach = function() {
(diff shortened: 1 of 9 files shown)


lib/link.js
```diff
@@ -151,17 +160,15 @@ Link.prototype._detached = function(frame) {
...
-  this.session._linksByName[this.policy.options.name] = undefined;
-  this.session._allocatedHandles[this.policy.options.handle] = undefined;
-  this.attached = false;
-  this.emit(Link.Detached, { closed: frame.closed, error: frame.error });
+  delete this.session._linksByName[this.policy.options.name];
+  delete this.session._allocatedHandles[this.policy.options.handle];
 
-  // @todo: Session should listen for Link.Detached
-  this.session._linkDetached(this);
+  this.emit(Link.Detached, { closed: frame.closed, error: frame.error });
```
noodlefrenzy: It seems like session could potentially listen for Detached, _or_ pass in some callback in createLink, to do this `delete this.session.XXX` work. Thoughts?

lib/sender_link.js
```diff
@@ -89,7 +89,7 @@ SenderLink.prototype.send = function(msg, options) {
       }
     };
 
-    if (!self.attached || !self.canSend()) {
+    if (self.state() !== 'attached' || !self.canSend()) {
```
noodlefrenzy: we should move `self.state() === 'attached'` check into `canSend()`

lib/link.js
mbroadst: Like this: https://github.com/noodlefrenzy/node-amqp10/blob/link-on-attach-handler/lib/session.js#L222 ?

lib/sender_link.js
mbroadst: totally

lib/link.js
noodlefrenzy: :) Yes

PR #141, opened 2015-08-03
feat(unique-link-names): use uniquely generated link names

mbroadst:
This follows more closely to what QPID is doing with their messaging framework, in that link names by default are generated based on their address (and in our case link creation properties). A hash of the address and options is used to uniquely identify an attempt to recreate an existing receiver. The user is still able to override this behavior by passing in a "name" parameter to the link creation "options" object (verified in the unit tests)

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -117,12 +117,16 @@ function Session(conn) {
   this.mapped = false;
   this.remoteChannel = undefined;
   this._allocatedHandles = {};
-  this._linksByName = {};
-  this._linksByRemoteHandle = {};
   this._deliveryTag = 1;
 
   this._senderLinks = {};
   this._receiverLinks = {};
+
+  this._linksByName = {};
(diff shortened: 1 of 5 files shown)


lib/session.js
```diff
@@ -117,13 +117,17 @@ function Session(conn) {
...
   this.remoteChannel = undefined;
   this._allocatedHandles = {};
-  this._linksByName = {};
-  this._linksByRemoteHandle = {};
   this._deliveryTag = 1;
 
   this._senderLinks = {};
   this._receiverLinks = {};
 
+  this._linksByName = {};
```
noodlefrenzy: Since we're only using these to resolve incoming attach frames, perhaps we can rename it something like _attachPendingLinks and `delete` the entry when we receive the attach. Right now it seems like we're double-book-keeping with `_senderLinks`/`_receiverLinks`.

PR #143, opened 2015-08-03
feat(default-subject): allow specifying subjects in link names

mbroadst:
This patch adds support for defining a subject in the name of a link during link creation. On the receiver side this adds a filter for the subject, and on the sender side this adds a default subject to the message properties (if one does not currently exist).

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -192,17 +192,21 @@ AMQPClient.prototype.createSender = function(address, options) {
     throw new Error('Must connect before creating links');
   }
 
-  address = address || this._defaultQueue;
+  address = u.parseLinkAddress(address || this._defaultQueue);
   options = options || {};
 
-  var linkName = u.linkName(address, options),
+  var linkName = u.linkName(address.name, options),
       linkPolicy = u.deepMerge({
         options: {
           name: linkName,
           source: { address: 'localhost' },
-          target: { address: address }
+          target: { address: address.name }
         }
       }, this.policy.senderLink);
+
+  if (!!address.subject && this.policy.defaultSubjects) {
+    linkPolicy.defaultSubject = address.subject;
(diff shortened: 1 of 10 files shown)


lib/amqp_client.js
```diff
@@ -192,18 +192,22 @@ AMQPClient.prototype.createSender = function(address, options) {
     throw new Error('Must connect before creating links');
   }
 
-  address = address || this._defaultQueue;
+  address = u.parseLinkAddress(address || this._defaultQueue);
```
noodlefrenzy: We'd talked about not calling parseLinkAddress if defaultSubjects is false, because it will do the wrong thing.

mbroadst: I wasn't sure what the EH/SB spec was for this. If slashes aren't allowed for queue names this should "just work TM"

noodlefrenzy: They do - receivers for EH typically look like _ehname_/ConsumerGroups/$default/Partitions/1

mbroadst: oh bummerama, okay I'll switch it up right quick

mbroadst:
@noodlefrenzy how's that? An additional point, it _might_ be a little confusing for readability that the top level feature scope is called `defaultSubjects`, and the senderLink level scope is called `defaultSubject`.. but I'm not sure otherwise where is best to place that.

noodlefrenzy:
I know (re: `defaultSubjects` vs. `defaultSubject`) but I'm not sure what to do there. Both make sense.  Perhaps change the option to `useDefaultSubjects` or `inferSubjectsFromAddress` or something?

Anyway, your change looks good - sorry it took me so long to respond, for some reason I keep expecting Github's tab-bar to update when a new message comes in, and it keeps disappointing me.

PR #147, opened 2015-08-06
Keep session links

noodlefrenzy:
I was closer to right than you were - it was more than just not setting session to null.

- Remove reattach logic from client
- Stop setting session to null on disconnect
- Stop session from clearing link state on detach
- Add reattach policies for link, akin to client's reconnect policies
- Ensure link doesn't attempt reattach if it's detached by our side
- Add a "forceDetach" to force existing links into a detached state if the session/connection is killed

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -74,7 +74,6 @@ function AMQPClient(policy) {
   this._connection = null;
   this._session = null;
 
-  this._reattach = {};
   this._reconnect = null;
   if (!!this.policy.reconnect) {
     this._timeouts = u.generateTimeouts(this.policy.reconnect);
@@ -224,7 +223,6 @@ AMQPClient.prototype.createSender = function(address, options) {
       link._onAttach.push(attachPromise);
     };
 
-    self._reattach[linkName] = attach;
     attach();
   });
 };
@@ -297,7 +295,6 @@ AMQPClient.prototype.createReceiver = function(address, options) {
       link._onAttach.push(attachPromise);
     };
 
-    self._reattach[linkName] = attach;
     attach();
   });
 };
@@ -329,13 +326,11 @@ AMQPClient.prototype.disconnect = function() {
 
 AMQPClient.prototype._clearConnectionState = function(saveReconnectDetails) {
   this._connection = null;
-  //this._session = null;
 
   // Copy from original to avoid any settings changes "sticking" across connections.
   this.policy = u.deepMerge(this._originalPolicy);
 
   if (!saveReconnectDetails) {
-    this._reattach = {};
     this._reconnect = null;
   }
 };
@@ -350,7 +345,6 @@ AMQPClient.prototype._newSession = function(conn) {
 };
 
 AMQPClient.prototype._preventReconnect = function() {
-  this._reattach = {};
   this._reconnect = null;
 };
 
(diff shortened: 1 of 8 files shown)


lib/link.js
```diff
@@ -100,11 +107,13 @@ Link.prototype.attach = function() {
 
 Link.prototype.detach = function() {
   var self = this;
+  this._timeouts = undefined; // Disable any re-attachment policy.
   var detachPromise = new Promise(function(resolve, reject) {
     var onError = function(err) { reject(err); };
     self.once(Link.ErrorReceived, onError);
     self.once(Link.Detached, function(info) {
       self.removeListener(Link.ErrorReceived, onError);
+      self.session._removeLink(self);
```
mbroadst: something feels wrong about this, shouldn't the session listen to Link.Detach events itself and remove the links from there? I guess this is better stated as: why is Session's internal state something Link cares about?

lib/link.js
```diff
@@ -164,10 +182,30 @@ Link.prototype._detached = function(frame) {
...
   this.emit(Link.Detached, { closed: frame.closed, error: frame.error });
   this._resolveAttachPromises(frame.error ? frame.error : 'link closed');
+
+  var self = this;
+  if (!self._shouldReattach()) return;
+
+  if (!self._timeouts.length) self._timeouts = u.generateTimeouts(self.policy.reattach);
+
+  setTimeout(function() {
+    // @todo: see _attemptReattach below for more detail, but bail here
```
mbroadst: no need for this comment

lib/link.js
noodlefrenzy: I struggled with this. Basically a session should only remove the link when it's given up on itself (either because we called detach or because it's run out of retries). The link is the one who knows when it's given up - and that _doesn't_ coincide with a Detach event. So either the session needs to interrogate the link on whether it might still be trying to attach, or the link needs to tell the session it's done. I chose the latter, but I can see why the former makes sense given our separation of concerns - I can do that.

mbroadst: Unless I'm reading this incorrectly the code as it stands still removes the link every Link.Detach though right? Nothing different than the existing Link.Detach in Session

noodlefrenzy: No, it removes it _only_ when the detach comes from our side. When the detach is instigated by the server, it leaves it alone.

lib/amqp_client.js
```diff
@@ -133,21 +132,26 @@ AMQPClient.prototype.connect = function(url) {
...
+          debug('unmapped');
+        });
+
+        self._session.on(Session.ErrorReceived, function(e) {
+          debug('session error: ', e);
+          self.emit(AMQPClient.ErrorReceived, e);
+        });
+      }
+
       self._session.on(Session.Mapped, function(s) {
```
mbroadst: I think we need to unregister this listener (existing bug)

lib/amqp_client.js
```diff
@@ -133,21 +132,26 @@ AMQPClient.prototype.connect = function(url) {
     self._connection.on(Connection.Connected, function(c) {
       debug('connected');
       self.emit(AMQPClient.ConnectionOpened);
-      self._session = self._newSession(c);
+      if (self._session) {
+        debug('session already exists, re-using');
+        self._session.connection = self._connection;
+      } else {
+        self._session = self._newSession(c);
+        self._session.on(Session.Unmapped, function(s) {
```
mbroadst: can't these two eventemitter callbacks just be moved to `_newSession`?

mbroadst: as an additional cleanup suggestion (while we're in this code here), it looks like we can move at least the Connection on('errorReceived') below to `_newConnection` as well maybe

PR #148, opened 2015-08-11
Harden heartbeat

noodlefrenzy:
Make sure the interval gets cleared, make sure it doesn't try and send if the socket (`client`) isn't set. Some windows makefile tweaks along the way.


mbroadst:
LGTM :+1:

PR #150, opened 2015-08-11
feat(ReceiverLink): add Released and Modified outcomes

mbroadst:
* Add two new disposition outcomes to ReceiverLink: Released and Modified
  * Share common code between all disposition methods, allowing a user to optionally directly use the   `settle` method as well
  * Documentation!


noodlefrenzy:
LGTM. I also like the comment format.

Issue #149, opened 2015-08-11
unify 'options' handling for sender and receiver links


mbroadst:
Both [AMQPClient.prototype.createSender](https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/amqp_client.js#L180) and [AMQPClient.prototype.createReceiver](https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/amqp_client.js#L232) take `options` as a second parameter, however, it's a bit confusing to the user what those options actually are. I'm opening this issue so we can discuss how best to move forward with these in v2.0.0.

Currently each link accepts options with the following criteria:

**ReceiverLink** 
  * `filter` if filter exists, and filter is an array with a map as the first entry (<-- incredibly specific), then it is set as the source filter on the receiver link
  * `policy` if policy exists, then it is deepMerged onto the existing policy (made up from the Client's receiverLinkPolicy, as well as options specific to the address passed in). 

**SenderLink**
  * `options` aren't used at all, other than to create a unique identifier for the linkName

So the question is: what _should_ `options` actually be for both link types?

noodlefrenzy:
Good summary. Are we still thinking of having a SimpleClient? If so, we could potentially punt on a top-level filter and just have options actually _be_ `policy` overrides, and those who want to specify filters need to pass in
```
{ 
  options: {
    source: {
      filter: ...
} } }
```

mbroadst:
When using the library the other day it seemed to me that `options` should be `attach frame options` (aka `link creation options`), which I totally understand to be restrictive and perhaps too low-level. On the other hand, this _felt_ a little awkward:

```
client.createReceiver(address, {
  policy: {
    options: {
      receiverSettleMode: c.receiverSettleMode.settleOnDisposition
    },
    creditQuantum: 1
  }
});
```

noodlefrenzy:
That would make it more painful for the client, but much more straightforward about what should be in there, and where.

noodlefrenzy:
The other option would be to only allow top-level policy settings in the policy and not attach-frame options, and name attach-frame options something like `attach`, and still allow filter, so it'd look something like:
```
{
  policy: { creditQuantum: 1},
  attach: { receiverSettleMode: ... },
  filter: ...
}
```

mbroadst:
[...]

PR #151, opened 2015-08-12
refactor(link-options): standardize link options for link creation

mbroadst:
see #149 

includes:
  * renaming `options` to `policyOverrides`
  * renaming `senderLinkPolicy.options` and `receiverLinkPolicy.options` to `attach`

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -172,27 +172,27 @@ AMQPClient.prototype.connect = function(url) {
  * Creates a sender link for the given address, with optional link policy
  *
  * @method createSender
- * @param {String} [address]    An address to connect this link to
- * @param {Object} [options]    Policy used for creating the sender link
+ * @param {String} address              An address to connect this link to. If not provided will use default queue from connection uri.
+ * @param {Object} [policyOverrides]    Policy overrides used for creating this sender link
  *
  * @return {Promise}
  */
-AMQPClient.prototype.createSender = function(address, options) {
+AMQPClient.prototype.createSender = function(address, policyOverrides) {
   if (!this._connection) {
     throw new Error('Must connect before creating links');
   }
 
   address = u.parseLinkAddress(address || this._defaultQueue, this.policy);
-  options = options || {};
+  policyOverrides = policyOverrides || {};
 
-  var linkName = u.linkName(address.name, options),
+  var linkName = u.linkName(address.name, policyOverrides),
       linkPolicy = u.deepMerge({
         options: {
           name: linkName,
           source: { address: 'localhost' },
           target: { address: address.name }
         }
-      }, options, this.policy.senderLink);
+      }, policyOverrides, this.policy.senderLink);
 
   if (!!address.subject && this.policy.defaultSubjects) {
     linkPolicy.defaultSubject = address.subject;
@@ -220,27 +220,27 @@ AMQPClient.prototype.createSender = function(address, options) {
  * used to listen for 'message' events.
  *
  * @method createReceiver
- * @param {String} [source]     Source of the link to connect to.  If not provided will use default queue from connection uri.
- * @param {Object} [options]    Link policy used for creating this receiver link
+ * @param {String} address              An address to connect this link to.  If not provided will use default queue from connection uri.
+ * @param {Object} [policyOverrides]    Policy overrides used for creating this receiver link
  *
  * @return {Promise}
  */
-AMQPClient.prototype.createReceiver = function(address, options) {
+AMQPClient.prototype.createReceiver = function(address, policyOverrides) {
   if (!this._connection) {
     throw new Error('Must connect before creating links');
   }
 
   address = u.parseLinkAddress(address || this._defaultQueue, this.policy);
-  options = options || {};
+  policyOverrides = policyOverrides || {};
 
-  var linkName = u.linkName(address.name, options),
+  var linkName = u.linkName(address.name, policyOverrides),
       linkPolicy = u.deepMerge({
         options: {
           name: linkName,
           source: { address: address.name },
           target: { address: 'localhost' }
         }
-      }, options, this.policy.receiverLink);
+      }, policyOverrides, this.policy.receiverLink);
 
   // if a subject has been provided then automatically set up a filter to match
   // on that subject.
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -169,11 +169,11 @@ AMQPClient.prototype.connect = function(url) {
 };
 
 /**
- * Creates a sender link for the given address, with optional link options
+ * Creates a sender link for the given address, with optional link policy
  *
  * @method createSender
- * @param {string} [address]    An address to connect this link to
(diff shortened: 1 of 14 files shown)


lib/amqp_client.js
```diff
@@ -169,11 +169,11 @@ AMQPClient.prototype.connect = function(url) {
 };
 
 /**
- * Creates a sender link for the given address, with optional link options
+ * Creates a sender link for the given address, with optional link policy
  *
  * @method createSender
- * @param {string} [address]    An address to connect this link to
- * @param {*} [options]         Options used for creating the link
+ * @param {String} [address]    An address to connect this link to
```
noodlefrenzy: My understanding was that `[params_in_brackets]` were being marked optional, and address is non-optional.

lib/amqp_client.js
```diff
@@ -169,11 +169,11 @@ AMQPClient.prototype.connect = function(url) {
...
+ * @param {Object} [options]    Policy used for creating the sender link
```
noodlefrenzy: can we actually change the param name to senderPolicy or policyOverrides?

lib/amqp_client.js
```diff
@@ -215,17 +215,13 @@ AMQPClient.prototype.createSender = function(address, options) {
...
+ * promise returned resolves to a link that is an EventEmitter, which can be
+ * used to listen for 'message' events.
  *
  * @method createReceiver
- * @param {string} [source]     Source of the link to connect to.  If not provided will use default queue from connection uri.
- * @param {Object} [options]    Options used for link creation
- * @param options.filter        Filter used in connecting to the source.  See AMQP spec for details, and your server's documentation
- *                              for possible values.  node-amqp-encoder'd maps will be translated, and simple maps will be converted
- *                              to AMQP Fields type as defined in the spec.
+ * @param {String} [source]     Source of the link to connect to.  If not provided will use default queue from connection uri.
```
noodlefrenzy: Ditto the above two comments.

lib/amqp_client.js
```diff
@@ -215,17 +215,13 @@ AMQPClient.prototype.createSender = function(address, options) {
...
+ * @param {Object} [options]    Link policy used for creating this receiver link
```
noodlefrenzy: I see you decided not to add an explicit filter param.

test/integration/qpid/disposition.test.js
```diff
@@ -69,12 +69,10 @@ describe('Disposition', function() {
         return Promise.all([
           test.broker.initialize(),
           test.client.createReceiver(queueName, {
-            policy: {
-              options: {
-                receiverSettleMode: c.receiverSettleMode.settleOnDisposition
-              },
-              creditQuantum: 1
-            }
+            options: {
```
noodlefrenzy: This does look nicer.

test/unit/test_amqpclient.js
```diff
@@ -106,7 +106,7 @@ describe('AMQPClient', function() {
 
       return client.connect(mock_uri)
         .then(function () {
-          return client.createSender(queue, { name: 'queue_TX' });
+          return client.createSender(queue, { options: { name: 'queue_TX' } });
```
noodlefrenzy: These... less so. Oh well.

noodlefrenzy: I feel kind of like calling the key _inside_ the link policy `options` is a mistake, and it should be `attachParams` or something.

lib/amqp_client.js
mbroadst: ah yeah totally, fixing

lib/amqp_client.js
mbroadst: yeah I was just going to name it `policy`, but then wondered if it would be confusing having a `policy` and `this.policy`. `linkPolicy` also appealed to me, however, we're already using that in the method itself.

Should we:
  * just call it `policy`
  * call it `senderPolicy`
  * rename `this.policy` => `this.clientPolicy`, call it `policy`
  * <something else>

lib/amqp_client.js
mbroadst: It's just a start :) I figured it was better to have some code up on a branch to track the conversation.

test/unit/test_amqpclient.js
mbroadst: yeah I definitely agree here. `attach`, `attachParams`, `attachOptions`, any preference?

lib/amqp_client.js
noodlefrenzy: Well, what it _really_ is are `linkPolicyOverrides` or `senderPolicyOverrides`, but if that's too long I'd understand. If so, `senderPolicy` is my preference.

lib/amqp_client.js
noodlefrenzy: That makes total sense.

lib/amqp_client.js
mbroadst: ah I forgot you included: `policyOverrides`, agree on that one?

test/unit/test_amqpclient.js
noodlefrenzy: I personally think just `attach` would be fine, since we're already contained inside a policy object so "attach policy" is implied. Note that this also means it could be `session.begin`, and it _also_ points out that I did a terrible separation on the `connection.options`

mbroadst: Also, we could maintain the name as a top level option (the relevant change for this is in `u.linkName`), but that does sort of add a special case when I _think_ the goal here is to standardize everything.. 

Writing `client.createReceiver(address, { name: 'something' })` feels very natural to me, whereas `client.createReceiver(address, { attachOptions: { name: 'something' } });` feels very awkward.. *shrug*

lib/amqp_client.js
noodlefrenzy: Works for me

test/unit/test_amqpclient.js
noodlefrenzy: Honestly, I'd be ok with special-casing name. It does feel like a) something people want to commonly set, and b) something we use internally far more than just in attaching.

lib/session.js
```diff
@@ -201,22 +201,23 @@ Session.prototype.begin = function(sessionPolicy) {
...
-  var policy = u.deepMerge(linkPolicy);
+  var policy = u.deepMerge(linkPolicy),
+      attachOptions = policy.attach || {};
 
-  policy.options.handle = this._nextHandle();
-  if (typeof policy.options.name === 'function')
-    policy.options.name = policy.options.name();
+  attachOptions.handle = this._nextHandle();
+  if (typeof attachOptions.name === 'function')
+    attachOptions.name = attachOptions.name();
```
noodlefrenzy: So we're not doing top-level names

mbroadst: actually we _are_, its just sort of hidden. In the createSender/createReceiver methods we override name with linkName, which supports a top level name :)

noodlefrenzy: It's magic!

PR #153, opened 2015-08-21
Remove handlers from dead connections

noodlefrenzy:
And destroy socket on error.

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -298,6 +298,7 @@ AMQPClient.prototype.disconnect = function() {
 };
 
 AMQPClient.prototype._clearConnectionState = function(saveReconnectDetails) {
+  this._connection.removeAllListeners();
   this._connection = null;
 
   // Copy from original to avoid any settings changes "sticking" across connections.
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -285,12 +285,13 @@ AMQPClient.prototype.disconnect = function() {
     debug('disconnecting');
     if (self._connection) {
       self._preventReconnect();
-      self._connection.on(Connection.Disconnected, function() {
-        self._connection = null;
+      var connection = self._connection;
+      self._clearConnectionState();
+      connection.once(Connection.Disconnected, function() {
+        debug('disconnected');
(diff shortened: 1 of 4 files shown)


test/integration/servicebus/queue.test.js
```diff
@@ -18,9 +18,10 @@ describe('ServiceBus', function() {
       test.client = new AMQPClient(Policy.ServiceBusQueue);
     });
 
-    afterEach(function() {
+    afterEach(function(done) {
       return test.client.disconnect().then(function() {
         test.client = undefined;
+        done();
```
mbroadst: hmm this shouldn't be needed, were you experiencing issues where it didn't complete this step in time?

lib/amqp_client.js
```diff
@@ -285,12 +285,13 @@ AMQPClient.prototype.disconnect = function() {
     debug('disconnecting');
     if (self._connection) {
       self._preventReconnect();
-      self._connection.on(Connection.Disconnected, function() {
-        self._connection = null;
+      var connection = self._connection;
+      self._clearConnectionState();
+      connection.once(Connection.Disconnected, function() {
+        debug('disconnected');
```
mbroadst: `connection = undefined` in here? You've transferred the connection object over to the local reference to it, but I'm not sure if the reference is guaranteed to be cleaned up (though atm I can't see why it wouldn't be).. better to be safe?

test/integration/servicebus/queue.test.js
noodlefrenzy: I was encountering another issue (the fix below to store a local copy of connection and attach the disconnect handler _after_ calling removeAllListeners fixed that), so technically I don't need this anymore. _But_ isn't it the correct way to do things since I'm calling an async method in here?

mbroadst: If you return the promise, the promise is guaranteed to be resolved as part of the mocha step. See "Working with Promises" over here: https://mochajs.org/

noodlefrenzy: That's magically delicious! Ok, I'll remove my `done()` call.

PR #155, opened 2015-09-01
refactor(encodeOrdering): make encodeOrdering part of prototype

mbroadst:
- also cleaned up described type extraction using methods for
   extraction and assignment from payload

### DIFF
--- a/lib/frames/attach_frame.js
+++ b/lib/frames/attach_frame.js
@@ -199,6 +199,12 @@ AttachFrame.prototype.Descriptor = {
   code: new Int64(0x00000000, 0x00000012)
 };
 
+AttachFrame.prototype.EncodeOrdering = [
+  'name', 'handle', 'role', 'senderSettleMode', 'receiverSettleMode',
+  'source', 'target', 'unsettled', 'incompleteUnsettled', 'initialDeliveryCount',
+  'maxMessageSize', 'offeredCapabilities', 'desiredCapabilities', 'properties'
+];
+
 AttachFrame.prototype._getPerformative = function() {
   var self = this;
   var performative = new DescribedType(AttachFrame, {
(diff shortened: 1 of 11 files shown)


lib/frames/attach_frame.js
```diff
@@ -199,6 +199,12 @@ AttachFrame.prototype.Descriptor = {
   code: new Int64(0x00000000, 0x00000012)
 };
 
+AttachFrame.prototype.EncodeOrdering = [
```
noodlefrenzy: Thanks for correcting another of my "Baby's first Javascript project" mistakes.

lib/utilities.js
```diff
@@ -6,7 +6,8 @@ var _ = require('lodash'),
     constants = require('./constants'),
     errors = require('./errors'),
 
-    uuid = require('uuid');
+    uuid = require('uuid'),
+    utilities = module.exports = {};
```
noodlefrenzy: Hmm, neat trick

lib/utilities.js
```diff
@@ -293,7 +294,51 @@ module.exports.dispositionRange = function(message) {
...
+utilities.extractDescribedType = function(Type, value, defaults) {
+  defaults = defaults || {};
+
+  var result = {},
+      ordering = Type.prototype.EncodeOrdering;
+  for (var i = 0; i < value.length; i++) {
+    result[ordering[i]] = value[i];
+  }
+
+  if (defaults === undefined)
```
noodlefrenzy: defaults will never === undefined due to line 318 (ditto for method below)

lib/utilities.js
```diff
@@ -293,7 +294,51 @@ module.exports.dispositionRange = function(message) {
...
+    return _.isNull(source[key]) || _.isUndefined(source[key]);
+  });
+
+  _.assign(dest, propertiesToAssign);
+};
+
+/**
+ * Extract values for a DescribedType based on that types EncodeOrdering
+ */
+utilities.extractDescribedType = function(Type, value, defaults) {
```
noodlefrenzy: This method doesn't pull its own weight - if you were to pass `{}` in as the _"instance"_ to `assignDescribedTypeFromPayload` it would do the exact same thing.

PR #156, opened 2015-09-01
feat(SenderLink): accept a full message as the first param for send

mbroadst:
SenderLink.send currently only accepts a signature in the form  of
(body, messageOptions), this patch allows the user to send the
entire formed message as the first parameter.

**NOTE** depends on #155, this PR will be much smaller once that is accepted and this is rebased

### DIFF
--- a/lib/sender_link.js
+++ b/lib/sender_link.js
@@ -71,7 +71,15 @@ SenderLink.prototype.send = function(msg, options) {
 
   var self = this;
   return new Promise(function(resolve, reject) {
-    var message = new M.Message(options, self.policy.encoder(msg));
+    var message;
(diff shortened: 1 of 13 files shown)


lib/sender_link.js
```diff
@@ -71,7 +71,15 @@ SenderLink.prototype.send = function(msg, options) {
 
   var self = this;
   return new Promise(function(resolve, reject) {
-    var message = new M.Message(options, self.policy.encoder(msg));
+    var message;
+    if (_.isObject(msg) && msg.hasOwnProperty('body')) {
+      msg.body = self.policy.encoder(msg.body);
```
noodlefrenzy: side-effect. maybe instantiate the message, _then_ encode the body in message and not msg?

PR #158, opened 2015-09-02
Add eventhub integration test, fix up Makefile.

noodlefrenzy:


Makefile
```diff
@@ -1,7 +1,11 @@
+ifdef GREP
+	GREPARG = -g $(GREP)
+endif
+
 REPORTER ?= spec
-UNIT_TESTS = $(shell find ./test/unit -name "test_*.js")
-QPID_INTEGRATION_TESTS = $(shell find ./test/integration/qpid -name "*.test.js")
-SERVICEBUS_INTEGRATION_TESTS = $(shell find ./test/integration/servicebus -name "*.test.js")
+UNIT_TESTS = ./test/unit
+QPID_INTEGRATION_TESTS = ./test/integration/qpid
```
mbroadst: hm technically the idea with the `$(shell find..)` was that we wouldn't have to specify explicit paths, but rather just pick up anything below a top level dir. tbh if it works, I don't mind for now :smile:

noodlefrenzy: I'm not sure how mine is specifying explicit paths - it's using the same paths that yours had, I'm just letting Mocha do the hard work of finding the tests instead of us doing it.

test/integration/servicebus/eventhubs/client.test.js
```diff
@@ -0,0 +1,215 @@
...
+        test.client = undefined;
+      });
+    });
+
+    it('should connect, send, and receive a message', function(done) {
+      var msgVal = uuid.v4();
+      expect(config.senderLink, 'Required environment variables').to.exist;
+      test.client.connect(config.address)
+        .then(function() {
+          return Promise.all(_.range(config.partitionCount).
```
mbroadst: fwiw you can clean this up a _little_ I think by doing:
```
  .then(function() {
    return Promise.all([
      Promise.map(_.range(config.partitionCount), function(partition) {
        return test.client.createReceiver(config.receiverLinkPrefix + partition);
      },
      test.client.createSender(config.senderLink))
    ]);
  });
```

test/integration/servicebus/eventhubs/client.test.js
```diff
@@ -0,0 +1,215 @@
...
+        })
+        .then(function(links) {
+          var sender = links.pop();
+          _.each(links, function(receiver) {
+            receiver.on('message', function (message) {
+              expect(message).to.exist;
+              expect(message.body).to.exist;
+              // Ignore messages that aren't from us.
+              if (!!message.body.DataValue && message.body.DataValue === msgVal) {
+                done();
```
mbroadst: this is going to call done on the first message received, so if you have multiple partitions you're not testing exactly what you think you are I think

test/integration/servicebus/eventhubs/client.test.js
noodlefrenzy: I didn't see Promise.map, thanks for telling me!

test/integration/servicebus/eventhubs/client.test.js
noodlefrenzy: Actually, yes it is, but I can't blame you for not knowing. In Event Hubs, each message goes to a single partition, so the only way to make sure you get the message is to a) on sending, send to a specific partition (I'll be writing that test later, once I figure out how :)), or b) attach to all partitions.

PR #162, opened 2015-09-17
moved closure outside of for loop for variable independence.

robblovell:
If you put the createReceiver's .then clause inside the for loop, all the variables used in receiver.on end up with the value of the last partition in the bind clause.

Also, I couldn't get the filterOption to work, so I left it for a later refactor.

[Review on Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/162)


noodlefrenzy:
Good catch, thanks!

PR #167, opened 2015-09-30
Servicebus integration to travis

noodlefrenzy:
Finally actually enabling the eventhub/servicebus tests in Travis. Also, added another one for EH that tests out the time-based filter for receivers.

I'll be adding another one to make sure it actually ensures messages prior to that time don't come through, but that'll take a bit more work.

[Review on Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/167)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,14 +35,3 @@ script:
   - make coverage
 after_script:
   - make codeclimate-send
-
-env:
-  global:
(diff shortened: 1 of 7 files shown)


noodlefrenzy:
Ok, updated with _actual_ correct filters, and a test that ensures the filters are being honored by EH.

.travis.yml
```diff
@@ -35,14 +35,3 @@ script:
   - make coverage
 after_script:
   - make codeclimate-send
-
-env:
```
mbroadst: don't we need these guys?

mbroadst:
LGTM

PR #168, opened 2015-10-01
Add example for getting partition information from event hub.

noodlefrenzy:
[Review on Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/168)

### DIFF
--- a/examples/eventhub_management_request.js
+++ b/examples/eventhub_management_request.js
@@ -1,0 +1,98 @@
+//================================
+// EventHub Management test - takes in a JSON settings file
+// containing settings for connecting to the Hub:
+// - protocol: should never be set, defaults to amqps
+// - SASKeyName: name of your SAS key which should allow send/receive
+// - SASKey: actual SAS key value
+// - serviceBusHost: name of the host without suffix (e.g. https://foobar-ns.servicebus.windows.net/foobar-hub => foobar-ns)
+// - eventHubName: name of the hub (e.g. https://foobar-ns.servicebus.windows.net/foobar-hub => foobar-hub)
+//
+// Connects to the $management hub of the service bus, and sends a request to read properties for the given event hub
+// Dumps out the number of partitions, their IDs, and then quits.
+//================================
+
+'use strict';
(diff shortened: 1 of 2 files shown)


examples/eventhub_management_request.js
```diff
@@ -0,0 +1,98 @@
...
+}
+var sasName = settings.SASKeyName;
+var sasKey = settings.SASKey;
+var eventHubName = settings.eventHubName;
+var numPartitions = settings.partitions;
+
+var uri = protocol + '://' + encodeURIComponent(sasName) + ':' + encodeURIComponent(sasKey) + '@' + serviceBusHost;
+var managementEndpoint = '$management';
+
+var errorHandler = function(myIdx, rx_err) {
```
mbroadst: this doesn't seem to be used

examples/eventhub_management_request.js
```diff
@@ -0,0 +1,98 @@
...
+  console.warn('==> RX ERROR: ', rx_err);
+};
+
+var messageHandler = function (myIdx, msg) {
```
mbroadst: this either

examples/eventhub_management_request.js
```diff
@@ -0,0 +1,98 @@
...
+    client.disconnect().then(function() {
+      console.log('=== Disconnected ===');
+      process.exit(0);
+    })
+  });
+  sender.on('errorReceived', function (tx_err) {
+    console.warn('===> TX ERROR: ', tx_err);
+  });
+  var request = { body: "stub", properties: { messageId: 'request1', replyTo: rxName }, applicationProperties: { operation: "READ", name: eventHubName, type: "com.microsoft:eventhub" } };
+  sender.send(request).then(function (state) {
```
mbroadst: this would normally be like `return sender.send(request);` and the `then` would happen on the same level as the below `catch`

examples/eventhub_management_request.js
noodlefrenzy: Good call, thanks for catching my copy-pasta issues.

PR #173, opened 2015-10-11
refactor address parsing

mbroadst:
- Address parsing has been moved into the policies, and a new policy has been added: `QpidJava`
   - This policy differs from the `DefaultPolicy` only in that it optionally parses vhosts. 
 - Ability to specify a vhost during sasl init has been added, and a unit test was added for that as well
 - MockServer has been updated to support a new method: `setExpectedFrameSequence` so that we can test frame expectations from the client side in the mock server. Specifying `false` for a particular frame indicates that we don't care about it, so as to make reading which frame we _are_ testing for more readable.

[Review on Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/173)


noodlefrenzy:
Review status: 0 of 13 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[lib/frames/sasl_frame.js, line 217 \[r1\]](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/173#-K0RXpUPqKRva7bsmMdX)** ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/f94e836500f7a661c308201cb130294832e6dda2/lib/frames/sasl_frame.js#L217)):</sup>
I think there's a util method that does exactly this (u.coerce?)


---

<sup>**[lib/policies/default_policy.js, line 128 \[r1\]](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/173#-K0RZT2rHaCDEcV2gm2O)** ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/f94e836500f7a661c308201cb130294832e6dda2/lib/policies/default_policy.js#L128)):</sup>
I'd rather see `var userPass = parsedAddress.auth.split(':', 2);` and then have you use that (in theory SASL tokens _could_ have a colon).


---

<sup>**[lib/policies/qpid_java_policy.js, line 13 \[r1\]](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/173#-K0R_6g0U6ih4fXKk4BK)** ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/f94e836500f7a661c308201cb130294832e6dda2/lib/policies/qpid_java_policy.js#L13)):</sup>
See, if you used `split('/', 2)` here you wouldn't need the join below, I believe


---

<sup>**[lib/utilities.js, line 90 \[r1\]](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/173#-K0R_WvUFD7g4GgeE_zT)** ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/f94e836500f7a661c308201cb130294832e6dda2/lib/utilities.js#L90)):</sup>
Yay! Deleted code!


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/173)

PR #186, opened 2015-11-04
feat(messaging): support multi frame messages

mbroadst:


lib/frames/frame.js
```diff
@@ -179,8 +179,11 @@ AMQPFrame.prototype._writePayload = function(bufBuilder, options) {
   codec.encode(performative, bufBuilder);
   var payload = this._getAdditionalPayload();
   if (payload !== undefined) {
-    //debugAMQP('Encoding payload: ' + JSON.stringify(payload));
-    codec.encode(payload, bufBuilder);
+    if (payload instanceof Buffer) {
```
mbroadst: is this check relevant anymore now that we basically assume `Buffer`s for receive and send?

lib/frames/transfer_frame.js
```diff
@@ -247,4 +255,66 @@ TransferFrame.prototype._getAdditionalPayload = function() {
...
+ */
+TransferFrame.prototype.decodePayload = function() {
+  var message = new M.Message();
+  var body = [];
+  var curIdx = 0;
+  var possibleFields = {
+    'header': M.Header, 'footer': M.Footer, 'deliveryAnnotations': M.DeliveryAnnotations,
+    'annotations': M.Annotations, 'properties': M.Properties, 'applicationProperties': M.ApplicationProperties
+  };
+  var isData = function(x) { return x instanceof M.Data; };
```
mbroadst: not that I think its relevant for this PR, but I think these should be defined outside the function in the module, otherwise we're recreating this on _every_ `decodePayload` call

mbroadst:
LGTM, my only comment is on a piece of code that I'm likely to refactor with the optimization work I had been doing already.

lib/frames/frame.js
noodlefrenzy: I don't think it is, removing.

test/unit/client.test.js
```diff
@@ -73,24 +83,170 @@ describe('Client', function() {
...
+
+      // build our expected buffer segments
+      var messageData = new Array(2048).join('0');
+      var message = new M.Message({ body: messageData });
+      var codecBuffer = new Builder();
+      Codec.encode(message, codecBuffer);
+      var messageBuffer = codecBuffer.get();
+
+      var expected = [];
+      expected.push(messageBuffer.slice(0, 482));
```
noodlefrenzy: Oddly specific :) No way to factor these out so it's less "magic number-y"?

PR #187, opened 2015-11-05
 Added example for DEBUG=amqp10:client env variable

snobu:


snobu:
Oh noes i broke the build.

image

mbroadst:
well that's really strange, seems to be related to the secret SB/EH variables built into travis, I'm forcing a rerun to see if it was a fluke

mbroadst:
hmm no dice, we'll have to wait for @noodlefrenzy to wake up to take a look at that since he set it up (and actually knows the contents of those env variables). They may have just expired or something silly.

noodlefrenzy:
I don't allow forks to see those env variables (otherwise people could just submit PRs with tests that just dump the keys). I'll take a look at what would be required to disable those tests on forks - I'm fairly certain there's a travis environment variable I should be using there. In the meantime, I'll merge the PR as it looks good, thanks!

mbroadst:
oh, duh :smile:

PR #188, opened 2015-11-05
optimize parts of the codebase based on v8 traceopt passes

mbroadst:


lib/codec.js
```diff
@@ -173,106 +167,89 @@ Codec.prototype.decode = function(buf, offset, forcedCode) {
...
-    case 'boolean':
-      encoder = types.builders[forceType || 'boolean'];
-      return encoder(val, buf);
-
-    case 'string':
-      encoder = types.builders[forceType || 'string'];
-      return encoder(val, buf);
-
+  switch (typeof value) {
+    case 'boolean': return types.builders.boolean(value, buffer);
```
noodlefrenzy: Good catch with all the forceType cleanup.

lib/codec.js
```diff
@@ -173,106 +167,89 @@ Codec.prototype.decode = function(buf, offset, forcedCode) {
...
-        return types.builders.map(val, buf, this);
-      }
-      break;
+Codec.prototype._encodeObject = function(value, buffer) {
+  if (value instanceof DescribedType) {
+    buffer.appendUInt8(0x00);
+    this.encode(value.descriptor, buffer);
+    this.encode(value.getValue() || [], buffer);
+    return;
+  } else if (value.encodeOrdering && value.encodeOrdering instanceof Array) {
```
noodlefrenzy: don't need else clause due to return.

lib/codec.js
```diff
@@ -173,106 +167,89 @@ Codec.prototype.decode = function(buf, offset, forcedCode) {
...
+  if (value instanceof ForcedType) return this.encode(value.value, buffer, value.typeName);
+  if (value instanceof AMQPSymbol) return types.builders.symbol(value.contents, buffer, this);
 
-    default:
-      throw new errors.NotImplementedError('encode(' + type + ')');
+  var encoder;
+  if (value instanceof Int64) {
+    encoder = (value < 0) ? types.builders.long: types.builders.ulong;
+    return encoder(value, buffer, this);
+  }  else if (value.encode && (typeof value.encode === 'function')) {
```
noodlefrenzy: don't need else clause due to return.

lib/link.js
```diff
@@ -127,6 +127,7 @@ Link.prototype.detach = function() {
 Link.prototype.forceDetach = function() {
   debug('Force-detach for ' + this.name + '. Current state: ' + this.state());
   this.linkSM.forceDetach();
+  this._detached();
```
noodlefrenzy: I worry slightly here, since we're now emitting a Detached event when forceDetach is called. If re-attach is triggered when that happens, it'd be bad imho, but I don't remember the flow well enough to dissuade this fix. Perhaps keep frame as undefined instead of coalescing to {} below, and only emit the event if it's defined? Also, what does shouldReattach do in this case?

mbroadst: I'm not quite sure, some extra tests around this would help a lot. The issue is in a basic test that a forceDetach comes in and the links are never cleaned up (because `_detached`) isn't called.

lib/session.js
```diff
@@ -304,52 +304,68 @@ Session.prototype._processFrame = function(frame) {
...
-        senderLink._flowReceived(frame);
-      });
-    }
-  } else if (frame instanceof TransferFrame) {
-    if (frame.handle !== null && this._linksByRemoteHandle[frame.handle]) {
-      this._transferReceived(frame);
-      this._linksByRemoteHandle[frame.handle]._messageReceived(frame);
-    } else {
-      console.warn('received Transfer frame for unknown link(' + frame.handle + '): ' + JSON.stringify(frame));
+  switch (frame.constructor) {
```
noodlefrenzy: Magic. I like it.

lib/link.js
noodlefrenzy: Ok, I've looked through the code and think that this would clean up what needs to clean up, without causing problems:

```
Link.prototype._detached = function(frame) {
  if (this.remoteHandle !== undefined) {
    delete this.session._linksByRemoteHandle[this.remoteHandle];
    this.remoteHandle = undefined;
  }

  this._resolveAttachPromises((frame && frame.error) ? frame.error : 'link closed');

  if (!frame) return; // Cleaned up link on forced detach, ensure we don't try and reattach or send events.

  if (frame && frame.error) {
    this.emit(Link.ErrorReceived, frame.error);
  }

  this.emit(Link.Detached, { closed: frame.closed, error: frame.error });

  var self = this;
  if (!self.shouldReattach()) return;

  if (!self._timeouts.length) self._timeouts = u.generateTimeouts(self.policy.reattach);

  setTimeout(function() {
    self._attemptReattach();
  }, self._timeouts.shift());
};
```

mbroadst: hmm but if we don't emit `Link.Detached` then the session will never cleanup the link. Perhaps we should emit that signal with a `{ force: true }` in the event that the passed in frame is `undefined` or something?

PR #194, opened 2015-12-05
add support for AMQP over websockets for Azure IoT Hub

pierreca:
Here's a first stab at separating the transport layer from connection.js (meaning extracting the dependencies to net and tls in specific files) and adding websocket support to enable amqp over websockets for Azure IoT Hub

the websocket library used has the benefit of being entirely written in javascript, unlike other popular ones. the good thing about this change is that it makes it easier to change or add transport layers by just creating another client in the socket folder, and then figuring out in connection.js how to decide whether we should be using plain sockets, secure sockets (tls) or websockets.

I have run unit tests but not integration tests - I might need a little help doing this :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ If you find any issues, please report them via GitHub.
 ## Implementation Notes ##
 
 +   Using node's built-in net/tls classes for communicating with the server.
++   Using [nodejs-websocket](https://github.com/sitegui/nodejs-websocket) for Azure IoT Hub support of AMQP over WebSockets.
 
 +   Data from the server is written to a buffer-list based on [Rod Vagg's BL](https://github.com/rvagg/bl).
 
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -23,7 +23,7 @@ var EventEmitter = require('events').EventEmitter,
     HeartbeatFrame = require('./frames/heartbeat_frame'),
     OpenFrame = require('./frames/open_frame');
 
-    
+
 
 /**
  * Connection states, from AMQP 1.0 spec:
@@ -539,7 +539,8 @@ Connection.prototype._connect = function(address, sasl) {
       self.client.connect(address);
       break;
     default:
-    // TODO: Error
+      // this should never happen because it's checked in default_policy.js (in parseAddress)
+      throw new Error('invalid protocol: ' + address.protocol);
   }
   
   self.client.on('connect', function() {
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -2,9 +2,11 @@
 
 var EventEmitter = require('events').EventEmitter,
     fs = require('fs'),
-    net = require('net'),
-    tls = require('tls'),
     util = require('util'),
+    
+    WebSocketClient = require('./sockets/websocketclient.js'),
+    BasicSocketClient = require('./sockets/basicsocketclient.js'),
+    TlsSocketClient = require('./sockets/tlssocketclient.js'),
 
     debug = require('debug')('amqp10:connection'),
     BufferList = require('bl'),
@@ -21,7 +23,7 @@ var EventEmitter = require('events').EventEmitter,
     HeartbeatFrame = require('./frames/heartbeat_frame'),
     OpenFrame = require('./frames/open_frame');
 
-
+    
 
 /**
  * Connection states, from AMQP 1.0 spec:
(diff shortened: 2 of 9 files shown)


mbroadst:
@pierreca very cool work, thanks for the contribution! I think we can and should definitely get this in, but I have a few comments (and I'm sure @noodlefrenzy has thoughts as well):

amqp10 was originally designed to be a pure javascript implementation, avoiding building native components at all costs (there was an initial module that built against apache qpid's proton that was the impetus for this module iirc). Having said that, your changes here would bring in a hard requirement to build a native module (I believe `ws` is in there somewhere, below `nodejs-websocket`), so we'd be breaking one of our core rules there. That doesn't mean that we shouldn't think of a way to get this in, so this is a great first step. 

My proposal would be to make a new module with this work (`amqp10-ws`, `amqp10-transport-ws`? etc), and to "plug it in" to amqp10, so people who don't need websockets don't need to pay the penalty of downloading and building `ws`. At this point I think the conversation drives towards what a proposal for such a plugin system would look like, should it just be for transport? it looks like it will also affect the url parsing (support for `wss` in the protocol there). I imagine something like:

```
var amqp = require('amqp10');
amqp.use(require('amqp10-ws'));

var client = amqp.Client();
client.connect('wss://<something>')
  .then(function() { ... });
```

Thoughts?

```

**Comment by noodlefrenzy** [comment]
Huh, crap, didn't realize this depended on native code. I know Node has, what, 7 websocket libraries? Any chance there's one that doesn't have native code?

Another option, similar to what @mbroadst is suggesting, is a protocol->provider map, with an ability to register protocol providers in amqp_client. So e.g. the code might look like:

```
var amqp = require('amqp10');
require('amqp10-ws').register(amqp);

var client = amqp.Client();
client.connect('wss://<something>')
  .then(function() { ... });
```

With `register` implemented as:

```
register = function(client) {
  client.registerProtocolHandler('wss', this);
};
```

The only reason I'm thinking in that direction is that `use` just seems a bit too generic.

pierreca:
Mike and I had this conversation about using library with no native dependency, and from what I could understand (I'm kinda new to this) nodejs-websocket (this one: https://github.com/sitegui/nodejs-websocket) had no dependencies at all (cf package.json). I did try to avoid ws because of that... What did I miss that makes you say it has a dependency on ws?
This being said, I think it's a good idea to implement a better decoupling anyway. it'd be nice if we could have a provider system that allow users to choose what websocket library to use anyway.

mbroadst:
@pierreca ah you're totally right my apologies, things using websockets generally use the `ws` package for its speed. It looks like `nodejs-websocket` indeed uses pure javascript to get the job done (perhaps at a performance penalty?).

That satisfies my complaints from earlier, but I'm still marginally interested in the idea of "transport plugins." I think it would be nice to encapsulate the testing in a separate module, and you could theoretically use `ws` if building a binary protocol wasn't an issue for you. I'll try to squeeze in a talk with @noodlefrenzy when he's up :smile:

noodlefrenzy:
Whew, good, I _thought_ he'd said it didn't have a native dependency.

So here's where I stand on the plugin idea:

1. I like mine better than yours, short-term, because it seems focused and easy to do.
1. I like yours better than mine, long-term, because it would potentially allow plugins for a variety of features, and could replace some of the more awkward parts of the policy concept. I just don't understand yet how we would get it to work in a manner that made sense to both clients and implementors, without building a "framework" and that's the last thing I want to do.
1. I would be willing to do the code changes required to move from mine to yours if and when we ever figure that shit out.

@pierreca do you want to take a first stab at the protocol plugin? If not, I can probably get to it tomorrow.

mbroadst:
@noodlefrenzy sounds good to me. On that note I have a few review-related questions/observations:
 * should we call them "tranports" rather than "clients" (as we might want to support something crazy, like zeromq or whatever) 
 * IMHO each "transport" should inherit from an abstract base (`AbstractTransport` or whatever), so we can get a feel for expected behaviors (this would also serve as the interface to code against for transport plugins)
  * Internal variable for `Client` would in turn be renamed `this._client` to `this._transport` (or socket, whatever you guys want to name these "sparkle ponies")
  * As you mentioned, we'd just have `Client.prototype.registerTransport` 

I think that makes the work pretty manageably split. If someone wanted to tackle the refactorings in-place (as in inside the codebase as the PR is currently split up), I would be totally willing to do the work of making the external modularity work. I suppose it's just based on everyone's schedules at this point :smile:

noodlefrenzy:
@mbroadst yes, transports makes sense, as does the abstract base. I'm in a conference all day today and part of tomorrow, so earliest I would be able to help is tomorrow afternoon. If no-one else tackles it before then, I'll take it on.

pierreca:
Happy to try and implement this tomorrow, @noodlefrenzy I can start and we can meet later to review/push. One clarification though, I assume you would keep net and tls in the amqp10 library to provide basic transport functionality built-in, right?

noodlefrenzy:
@pierreca - exactly (re: net/tls), they would just implement whatever AbstractTransport we set up.

pierreca:
ok, there's a start here: #195

pierreca:
no need for this PR anymore, replaced by #195

PR #195, opened 2015-12-08
Separate transport from connection - Create the basics for transport injection

pierreca:
Here's a new set of changes to abstract the transport layer from the connection layer and allow new transport injection for future work with websockets, per the discussion in #194 .

Unit tests should pass but I haven't tested or ran this code with any sample code, at this point it's just a base for discussion (make sure I'm going in the right direction)


lib/transport/abstract-transport.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+var util = require("util");
+var EventEmitter = require('events');
```
noodlefrenzy: EventEmitter = require('events').EventEmitter

lib/transport/net-transport.js
```diff
@@ -0,0 +1,55 @@
...
+var AbstractTransport = require('./abstract-transport.js');
+var net = require('net');
+var debug = require('debug')('amqp10:connection:nettransport');
+
+var NetTransport = function () {
+	AbstractTransport.call(this);
+	this._transport = null;
+};
+
+NetTransport.prototype = Object.create(AbstractTransport.prototype);
```
noodlefrenzy: Curious why you're not using util.inherits.

lib/transport/abstract-transport.js
noodlefrenzy: Oh, hey, looking at latest docs looks like that's ok now. Not sure how well it does on previous versions, though. @mbroadst - any ideas when .EventEmitter stopped being required? Looks like after 0.10 maybe?

lib/transport/net-transport.js
```diff
@@ -0,0 +1,55 @@
...
+
+NetTransport.prototype.register = function (client) {
+	client.registerProtocolHandler('amqp', this);
+};
+
+NetTransport.prototype.connect = function (address) {
+		debug('Connecting to ' + address.host + ':' + address.port + ' via straight-up sockets');
+		this._transport = net.connect({ port: address.port, host: address.host });
+		
+		this._transport.on('connect', function() { this.emit('connect'); }.bind(this));
```
noodlefrenzy: Why .bind(this) instead of a closure on self=this? (e.g. `var self=this;`...)

lib/transport/net-transport.js
```diff
@@ -0,0 +1,55 @@
...
+	}
+};
+
+NetTransport.prototype.destroy = function() {
+	if (this._transport) {
+		this._transport.destroy();
+		this._transport = null;
+		this.removeAllListeners();
+	} else {
+		throw new Error('Socket not connected');
```
noodlefrenzy: I'm thinking we should make destroy idempotent. Remove this error and if someone calls destroy when `_transport` is already nulled out that should be ok.

lib/transport/tls-transport.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+var AbstractTransport = require('./abstract-transport.js');
+var tls = require('tls');
+var debug = require('debug')('amqp10:connection:tlstransport');
+
+var TlsTransport = function() {
```
noodlefrenzy: Ditto to the comments on NetTransport.

noodlefrenzy:
Other than the comments on Net/TlsTransport, looks good. I like the usage pattern.

lib/transport/abstract-transport.js
mbroadst: yeah it was added around 0.12.x, I'd say to be safe (since we do claim to support 0.10.x I guess) just stick with `require('events').EventEmitter`. It's an alias in the newer code anyway

lib/transport/abstract-transport.js
```diff
@@ -0,0 +1,40 @@
...
+
+var AbstractTransport = function () {
+	if (this.constructor === AbstractTransport) {
+		throwBecauseAbstract();
```
mbroadst: it's a little more work, but I'd prefer the errors thrown to be explicit. For instance, here it would be `cannot instantiate AbstractTransport`, below it would be `Subclass must implement 'write'`. Since we aren't going to export this class, rather just duck type outside implementations, implementors will probably run into these and more direction would be useful

lib/transport/net-transport.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+var AbstractTransport = require('./abstract-transport.js');
+var net = require('net');
+var debug = require('debug')('amqp10:connection:nettransport');
```
mbroadst: `var debug = require('debug')('amqp10:transport:net');` ?

lib/transport/net-transport.js
mbroadst: +1

lib/transport/net-transport.js
```diff
@@ -0,0 +1,55 @@
...
+
+var NetTransport = function () {
+	AbstractTransport.call(this);
+	this._transport = null;
+};
+
+NetTransport.prototype = Object.create(AbstractTransport.prototype);
+NetTransport.prototype.constructor = NetTransport;
+
+NetTransport.prototype.register = function (client) {
```
mbroadst: tbh I'm not really a fan of this indirection. I'd prefer to just straight up call `AmqpClient.registerTransport(require('some-transport-module'))`. Then again I'm not too too picky about it, just seems needless.

lib/transport/net-transport.js
```diff
@@ -0,0 +1,55 @@
...
+	client.registerProtocolHandler('amqp', this);
+};
+
+NetTransport.prototype.connect = function (address) {
+		debug('Connecting to ' + address.host + ':' + address.port + ' via straight-up sockets');
+		this._transport = net.connect({ port: address.port, host: address.host });
```
mbroadst: another nitpick: this class itself represents the transport, however what's returned from `net.connect` (and likely in all cases) is a `socket`, so maybe `this._socket`

lib/transport/net-transport.js
```diff
@@ -0,0 +1,55 @@
...
+		
+		this._transport.on('connect', function() { this.emit('connect'); }.bind(this));
+		this._transport.on('data', function(data) { this.emit('data', data); }.bind(this));
+		this._transport.on('error', function(err) { this.emit('error', err); }.bind(this));
+		this._transport.on('end', function() { this.emit('end'); }.bind(this));
+};
+
+NetTransport.prototype.write = function (data) {
+	if(this._transport) {
```
mbroadst: another nitpick (nitpicks galore!), I've made an effort in the code to keep the style more towards a "fail early and often" mostly to reduce brace level. Suggested form:

```
if (!this._transport) {
  throw new Error('Socket not connected');

this._transport.write(data);
```

lib/transport/net-transport.js
noodlefrenzy: Oh yeah, I like that better. Good call.

lib/transport/net-transport.js
noodlefrenzy: I like it, actually :) For some reason, I like that you're registering an instance of a transport on an instance of an AMQPClient, and that the transport is the one who knows (internally) what protocols it supports.

lib/transport/net-transport.js
noodlefrenzy: +1, good catch.

mbroadst:
Looks mostly good to me. A final comment on design:

Rather than have the transports live in a dictionary inside the `AMQPClient` itself, you might consider providing a `index.js` in the `transports` directory itself which acts as a singleton "transport provider", sort of like:

```
'use strict';
function TransportProvider() {
  this._transports = {};
}

TransportProvider.prototype.registerTransport = function(transport) {
   // NOTE: if we only cared about internal modules this would work, but maybe
   //             some other method could work for the duck-typed approach
   // if (! (transport instanceof AbstractTransport))
   //  throw new Error('object must be an abstract transport, or implement xyz');

   this._transports[transport.protocol] = transport;
};

module.exports = function() {
   var provider = new TransportProvider();

   // pre-register our "known" transports
   provider.registerTransport(require('./net-transport'));
   provider.registerTransport(require('./tls-transport'));
   return provider;
}
```
(note: you'll see that I added an extra `protocol` on the Transport objects themselves there)

I find the design a little bit cleaner, but ymmv!

noodlefrenzy:
Re: @mbroadst's comment - I'm not a big fan of `transport.protocol` mainly because I want the flexibility of a transport being able to register multiple protocols it can handle, which is why I was more of a fan of the double-dispatch method @pierreca had in there. Of course, that could just as easily be solved by having it be `transport.protocols` instead, so if you both decide you like Matt's design better, feel free :)

I do like the idea of a separate `TransportProvider` controlling the map, but I think I'd still prefer the double-dispatch method for registering.

lib/transport/abstract-transport.js
pierreca: I think we have to export this class, it's providing inheritance from EventEmitter

mbroadst: well that's one of the checks you can make during registration. I think you'll find it a far more agreeable design to not export the `AbstractTransport`, otherwise you'll have to add a circular dependency in your custom transport. This class can be used internally as the base class of the known transports for sure though

pierreca: I'm not sure there's a point in having an AbstractTransport class at all, if it's not used to enforce implementation outside of the amqp10 library. I understand the ugliness of the circular dependency, but it has its usefulness. if we're duck typing outside implementations, why not do it for ours too. after all, the built-in implementations are the safest

mbroadst: Well what I thought was particularly clever about your initial implementation was unifying the "interfaces" of the two existing transports, hence the idea for an `AbstractTransport class. Also, it's incredibly likely that we will be able to make use of the `AbstractTransport` for unit testing.

As far as exporting it goes: I'm not sure it makes sense to require the entirety of the `amqp10` module simply to implement a custom transport. Duck-typing may feel unsafe/funny/whatever, but it's sort of the law of the land in javascript/node world (perhaps a look at other similar cases would be illuminating here, maybe sails' waterline db backend or something as its highly modular).

I'm not too tied to any of it, just dropping my two cents

I'm amenable to either approach, I just felt that using the abstract for internal transports buys us some consistency/reliability cheaply.

pierreca:
@mbroadst, @noodlefrenzy, most of your comments were addressed with this new iteration. Let me know if you reach a decision on how to handle the protocol map in transport provider.

lib/transport/abstract-transport.js
pierreca: done :)

lib/transport/net-transport.js
pierreca: no reason. I'm fairly new to javascript, I'm doing inheritance how I read it was supposed to be done, I guess that was outside of the context of node and util.inherits

lib/transport/net-transport.js
pierreca: no particular reason, switched to the closure pattern instead.

lib/transport/net-transport.js
pierreca: done :)

lib/transport/abstract-transport.js
pierreca: no worries, I've removed AbstractTransport from the exports but left it there for the built-in transports. I'm asking because i'm new to all this :) I tried to find something to check for basic transport functionalities in TransportProvider.

lib/transport/net-transport.js
pierreca: done

lib/transport/net-transport.js
pierreca: you guys will have to decide on that one :) that's the last open question :)

lib/transport/net-transport.js
pierreca: done. _socket it is

lib/transport/net-transport.js
pierreca: Happy to adhere to your coding style, done :)

noodlefrenzy:
@pierreca, @mbroadst and I have talked and we've agreed to a compromise - we both like the factoring out from the client into its own `TransportProvider` class, but I've managed to get him to agree to your double-dispatch mechanism for registration. Make sense?

pierreca:
@noodlefrenzy  @mbroadst  ok, so as I understand it, it looks like the current state of the code would work, right? can you take another look at the code and see if you see anything else? This code works with the iothub samples and the unit tests pass, so I'm going to fix the jshint error that is making Travis fail and then how do you suggest we run integration tests?

lib/connection.js
```diff
@@ -523,19 +522,15 @@ Connection.prototype._connect = function(address, sasl) {
...
-  if (isSSL) {
-    var sslOptions = self._sslOptions || {};
-    sslOptions.port = self.address.port;
-    sslOptions.host = self.address.host;
-    self.client = tls.connect(sslOptions);
-    debug('Connecting to ' + self.address.host + ':' + self.address.port + ' via TLS');
-  } else {
-    self.client = net.connect({ port: self.address.port, host: self.address.host });
-    debug('Connecting to ' + self.address.host + ':' + self.address.port + ' via straight-up sockets');
+  self._transport = this._transportProvider.getTransportFor(address.protocol);
```
mbroadst: since this was implemented as a singleton, you just need to do:
`TransportProvider.getTransportFor(address.protocol)`

lib/connection.js
```diff
@@ -145,12 +143,13 @@ var EventEmitter = require('events').EventEmitter,
...
  */
-function Connection(connectPolicy) {
+function Connection(connectPolicy, transportProvider) {
   Connection.super_.call(this);
   var options = connectPolicy.options;
   u.assertArguments(options, ['containerId', 'hostname']);
   this.policy = connectPolicy;
-  this.client = null;
+  this._transport = null;
+  this._transportProvider = transportProvider;
```
mbroadst: no need to store transportProvider here

mbroadst:
@pierreca don't worry about those failing integration tests for SB/EH, we just don't export those environment variables for any branch other than `master`

lib/amqp_client.js
```diff
@@ -59,7 +60,8 @@ function AMQPClient(policy) {
   this.policy = u.deepMerge(this._originalPolicy);
   this._connection = null;
   this._session = null;
-
+  this._transportProvider = new TransportProvider();
```
noodlefrenzy: I don't think you need to store this, _or_ new it up, since it was implemented as a singletion. @mbroadst correct me if I'm wrong.

noodlefrenzy:
Hey @pierreca - I just had a mini panic attack about something. By making the `TransportProvider` a singleton (IMO the right decision), and registering instances of `Transport` classes to handle a given protocol, if we instantiate multiple `AMQPClient` classes they'll wind up sharing the same `Transport` instance.

For example, if we create two clients and have them both talk TLS, they'll share the same `TlsTransport` instance and one will clobber the other's socket.

**SO** I talked it over with Matt, and I think:
- `TransportProvider` remains a singleton, as implemented
- `AMQPClient` no longer stores `._transportProvider` or tries to new one up
- `TransportProvider.registerTransport` takes in a protocol, and a _factory_ for a `Transport` instance
- `TransportProvider.getTransportFor` invokes that factory, handing back a new `Transport` instance

Does that all make sense? I don't actually think it's that significant a change from what you have (at least I hope not), and I think it provides better isolation between the client and the transport, and better safety between clients.

lib/amqp_client.js
```diff
@@ -272,6 +273,17 @@ AMQPClient.prototype.disconnect = function() {
...
 };
 
+/**
+ * Registers a protocol and its associated transport object.
+ * 
+ * @method registerProtocolHandler
+ * @param {String}   protocol            The protocol to be used when communicating with the server.
+ * @param {Function} transportFactory  A factory method that instantiates a new transport object, which must be an implementation of AbstractTransport.
+ */ 
+AMQPClient.prototype.registerProtocolHandler = function (protocol, transportFactory) {
```
noodlefrenzy: I think we should remove this and expose TransportProvider from index.js.

lib/transport/index.js
```diff
@@ -0,0 +1,47 @@
...
+        throw new Error('transport should implement the \'' + member + '\' method.');
+      }
+   }
+   
+   if (!(transport instanceof EventEmitter))
+   {
+     throw new Error('transport should inherit from EventEmitter');
+   }
+
+   this._transports[protocol] = transport;
```
noodlefrenzy: The goal is to store the factory here...

lib/transport/index.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+TransportProvider.prototype.getTransportFor = function(protocol) {
+  if (!this._transports.hasOwnProperty(protocol))
+    throw new Error('invalid protocol: ', protocol);
+
+  return this._transports[protocol];
```
noodlefrenzy: ...and then use it down here, so every getTransportFor call gets a new instance. Otherwise multiple clients would share the same transport instance and stomp on each other.

lib/transport/net_transport.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+var AbstractTransport = require('./abstract_transport.js');
+var util = require("util");
```
noodlefrenzy: 'util' not "util"

There, I won the award for the most nitpicky code-review comment!

pierreca: I had to read that comment 3 times to see the difference. lol :)

noodlefrenzy:
I think I'm now all good with this. @mbroadst ?

mbroadst:
http://cdn.meme.am/instances/500x/57632994.jpg

PR #200, opened 2016-01-26
refactor types (step 1)

mbroadst:
This cleans up the way the types registry is created, also some new tests added and minor bugs fixed


lib/types.js
```diff
@@ -224,7 +226,7 @@ Types.prototype._arrayDecoder = function(width, buffer, codec) {
  * @param {Number} [width]      Should be 1 or 4.  If given, builder assumes code already written, and will ensure array is encoded to the given byte-width type.  Useful for arrays.
  * @private
  */
-Types.prototype._mapBuilder = function(map, bufb, codec, width) {
+function mapBuilder(map, bufb, codec, width) {
   if (typeof map !== 'object') {
     throw new errors.EncodingError('Unsure how to encode non-object as array');
```
noodlefrenzy: bug in original: should say "non-object as map"

PR #204, opened 2016-02-05
streamline described type codec

mbroadst:


lib/types/amqp_error.js
```diff
@@ -15,15 +14,11 @@ function AMQPError(condition, description, info) {
 
 util.inherits(AMQPError, DescribedType);
 
+AMQPError.prototype.Descriptor = { code: 0x1D, name: new AMQPSymbol('amqp:error:list') };
```
noodlefrenzy: do we need the `new AMQPSymbol` here?

lib/types/message.js
```diff
@@ -30,11 +29,7 @@ function Header(options) {
 
 util.inherits(Header, DescribedType);
 
-Header.prototype.Descriptor = {
-  name: new AMQPSymbol('amqp:header:list'),
-  code: new Int64(0x0, 0x70)
-};
-
+Header.prototype.Descriptor = { code: 0x70, name: new AMQPSymbol('amqp:header:list') };
```
noodlefrenzy: ditto here re: `new AMQPSymbol`

PR #208, opened 2016-02-16
the great type refactor

mbroadst:


lib/adapters/translate_encoder.js
```diff
@@ -73,13 +72,12 @@ var encoder = processor({
       for (var idx = 0; idx < vals.length; idx += 2) {
         var k = encoder(vals[idx]);
         var v = encoder(vals[idx + 1]);
-        if (k instanceof AMQPSymbol) {
+        result[k] = v;
```
noodlefrenzy: I think the reason I had this was that I couldn't have object keys (e.g. keys which where themselves objects). Apparently I was wrong?

mbroadst: yeah but now fields are encoded as `ForcedType('fields', value)`, and `AMQPSymbol` doesn't exist anymore. The encoding information exists from the schema of whatever is being encoded (frame, composite etc), so you'll never get anything but an assignable type here

noodlefrenzy: yeah, but that doesn't change the fact that you're doing `result[<instance of ForcedType>] = v;`

lib/codec.js
```diff
@@ -210,26 +207,17 @@ Codec.prototype._encodeObject = function(value, buffer) {
...
-    for (var _i = 0; _i < _len; ++_i) {
-      if (value[value.encodeOrdering[_i]] === undefined)
-        throw new errors.EncodingError('Encoding map as list failed, encodeOrdering[' +
-          value.encodeOrdering[_i] + '] field not found or set as undefined in map: ' +
-          JSON.stringify(value));
-      asList.push(value[value.encodeOrdering[_i]]);
-    }
-    return this.encode(asList, buffer);
+  // support for composite types
+  if (value.toDescribedType && typeof value.toDescribedType === 'function') {
```
noodlefrenzy: So much nicer.

lib/codec.js
```diff
@@ -182,27 +178,30 @@ Codec.prototype.encode = function(value, buffer, forceType) {
...
 
+Codec.prototype._encodeNumber = function(value, buffer) {
+  if (isNaN(value) || !isFinite(value)) {
+    throw new errors.EncodingError('unable to encode value: ', value);
+  }
+
+  if (Math.floor(value) - value !== 0) {
+    return types.double.encode(value, buffer);
+  } else if (value > 0) {
+    if (Math.abs(value) < 0x7FFFFFFF)
```
noodlefrenzy: no need for Math.abs, since we just proved it's positive.

noodlefrenzy: also, uint should in theory be able to encode up to 0xFFFFFFFF

lib/connection.js
```diff
@@ -405,18 +400,15 @@ Connection.prototype._tryReceiveHeader = function(header) {
 };
 
 Connection.prototype._tryReceiveFrame = function() {
-  try {
-    var frame = FrameReader.read(this._buffer);
-    if (frame === undefined) {
+  var frame = frames.readFrame(this._buffer);
```
noodlefrenzy: Honestly, this method `_tryReceiveFrame` is spurious and can be deleted in favor of just calling `frames.readFrame`

lib/frames.js
```diff
@@ -1,16 +1,32 @@
...
+
+frames.SaslFrame = function() {
+  frames.SaslFrame.super_.call(this, constants.frameType.sasl);
+};
+util.inherits(frames.SaslFrame, frames.Frame);
 
 frames.writeFrame = function(frame, stream) {
-  if (!(frame instanceof Frame)) {
+  if (!(frame instanceof frames.Frame)) {
     throw new errors.EncodingError('unknown frame type: ' + frame.type);
```
noodlefrenzy: honestly, might make sense to use JSON.stringify instead of `frame.type` here, since we already know it's not a Frame so probably doesn't have a type (or if it does, it's accidental).

lib/frames.js
```diff
@@ -26,17 +42,253 @@ frames.writeFrame = function(frame, stream) {
...
+    debug('Heartbeat frame: ' + sizeAndDoff.toString('hex'));
+    return;
+  }
+
+  var xHeaderSize = (doff * 4) - 8;
+  if (xHeaderSize > 0) {
+    var xHeaderBuf = buffer.slice(0, xHeaderSize);
+    buffer.consume(xHeaderSize);
+
+    // @todo Process x-header
```
noodlefrenzy: TODO: we should normalize our @todo messages :)

lib/frames.js
```diff
@@ -26,17 +42,253 @@ frames.writeFrame = function(frame, stream) {
...
+  if (buffer.length < 8) return undefined;
+
+  var sizeAndDoff = buffer.slice(0, 8);
+  var size = sizeAndDoff.readUInt32BE(0);
+  if (size > buffer.length) return undefined;
+  buffer.consume(8);
+
+  var doff = sizeAndDoff[4];
+  var frameType = sizeAndDoff[5];
+  if (frameType !== FrameType.AMQP && frameType !== FrameType.SASL) {
```
noodlefrenzy: I almost feel like we should move this below `buffer.consume(payloadSize)` - that way if we just get some unknown frame type, we still error, but they could potentially swallow that and continue receiving on the next frame.

mbroadst: sure, good idea

mbroadst: merged

lib/frames.js
```diff
@@ -26,17 +42,255 @@ frames.writeFrame = function(frame, stream) {
...
+  ]
+});
+
+frames.AttachFrame = defineFrame(FrameType.AMQP, {
+  name: 'attach', code: 0x12,
+  fields: [
+    { name: 'name', type: 'string', mandatory: true },
+    { name: 'handle', type: 'uint', mandatory: true },
+    { name: 'role', type: 'boolean', mandatory: true },
+    { name: 'sndSettleMode', type: 'ubyte', default: 2 },
```
noodlefrenzy: Can we use the constants for defaults here? I'm not a big magic-number fan.

lib/adapters/translate_encoder.js
mbroadst: Ah good point :) fixed

lib/frames.js
mbroadst: for sure, these were just direct parses from the spec. Also of note: if you look below here to `properties` I included a default of `{}` which is _not_ in the spec , but I included it to satisfy our existing tests. Think we should remove that?

noodlefrenzy: if {} converts to a null, then no it's ok.

mbroadst: iirc {} converts to: `0xc1, 0x01, 0x00`

lib/frames.js
```diff
@@ -26,17 +42,255 @@ frames.writeFrame = function(frame, stream) {
...
+});
+
+frames.DispositionFrame = defineFrame(FrameType.AMQP, {
+  name: 'disposition', code: 0x15,
+  fields:[
+    { name: 'role', type: 'boolean', mandatory: true },
+    { name: 'first', type: 'uint', mandatory: true },
+    { name: 'last', type: 'uint' },
+    { name: 'settled', type: 'boolean', default: false },
+    { name: 'state', type: '*' },
```
noodlefrenzy: how come this is 'delivery_state' in Transfer, but here is '*'?

lib/frames.js
```diff
@@ -26,17 +42,255 @@ frames.writeFrame = function(frame, stream) {
...
+    { name: 'resume', type: 'boolean', default: false },
+    { name: 'aborted', type: 'boolean', default: false },
+    { name: 'batchable', type: 'boolean', default: false }
+  ]
+});
+
+frames.DispositionFrame = defineFrame(FrameType.AMQP, {
+  name: 'disposition', code: 0x15,
+  fields:[
+    { name: 'role', type: 'boolean', mandatory: true },
```
noodlefrenzy: constant for default here too plz if possible, magic booleans are just as bad as magic numbers for fields like this (i.e. fields named 'role' of type 'boolean' - wtf, "yes, I'm a role")

lib/frames.js
```diff
@@ -26,17 +42,255 @@ frames.writeFrame = function(frame, stream) {
...
+frames.CloseFrame = defineFrame(FrameType.AMQP, {
+  name: 'close', code: 0x18,
+  fields: [
+    { name: 'error', type: 'error' }
+  ]
+});
+
+// SASL frames
+frames.SaslMechanismsFrame = defineFrame(FrameType.SASL, {
+  name: 'sasl_mechanisms', code: 0x40,
```
noodlefrenzy: I actively don't care, but feel morally obligated to point out the naming inconsistency between frame_names and fieldNames. Don't feel compelled to "fix" it.

lib/receiver_link.js
```diff
@@ -157,17 +156,18 @@ ReceiverLink.prototype._messageReceived = function(transferFrame) {
...
+    this._currentTransferFrame.payload.append(transferFrame.payload);
     curFrame = this._currentTransferFrame;
     this._currentTransferFrame = null;
   }
 
-  var message = curFrame.decodePayload();
+  var message = m.decodeMessage(curFrame.payload);
   // store deliveryId for later use
-  message._deliveryId = curFrame.deliveryId;
+  Object.defineProperty(message, '_deliveryId', { value: curFrame.deliveryId });
```
noodlefrenzy: I like this change.

lib/frames.js
mbroadst: well we could go with snake_case for field names as well, but the general argument here is: field names are idiomatic names for properties that will be used by developers when working with a frame, whereas the name of the frame is used to create the symbolic descriptor for the composite type (though I do believe it should be `sasl-mechanisms` here instead of using the _)

lib/frames.js
mbroadst: this is a good point. In fact above I utilize `requires` for `source` and `target` to ensure a certain type. What's `require`d here is a wrapper for all delivery states, to ensure the valid type is being applied. What's a little more troublesome is _why_ `delivery_state` even works above, as its essentially doing a `new ForcedType('delivery_state', value)`, and there is no such type. I'll need to look into this.

lib/sender_link.js
```diff
@@ -75,10 +73,15 @@ SenderLink.prototype.send = function(msg, options) {
   return new Promise(function(resolve, reject) {
     var message;
     if (u.isObject(msg) && msg.hasOwnProperty('body')) {
-      message = new M.Message(msg);
+      message = msg;
       message.body = self.policy.encoder(message.body);
     } else {
-      message = new M.Message(options, self.policy.encoder(msg));
+      if (u.isObject(options)) {
+        options.body = self.policy.encoder(msg);
```
noodlefrenzy: I realize this is a pre-existing bug, but should we protect against missing self.policy.encoder? We do against this.policy.decoder in receiver.

lib/session.js
```diff
@@ -290,29 +285,28 @@ Session.prototype._nextHandle = function() {
...
-    default:
-      debug('not yet processing frames of type: ' + frame.constructor.name);
-  }
+  debug('processing frame: ', frame);
+  if (frame instanceof frames.EndFrame) return this._processEndFrame(frame);
+  else if (frame instanceof frames.AttachFrame) return this._processAttachFrame(frame);
+  else if (frame instanceof frames.DetachFrame) return this._processDetachFrame(frame);
+  else if (frame instanceof frames.FlowFrame) return this._processFlowFrame(frame);
+  else if (frame instanceof frames.TransferFrame) return this._processTransferFrame(frame);
+  else if (frame instanceof frames.DispositionFrame) return this._handleDisposition(frame);
```
noodlefrenzy: We should just rename this `_processDispositionFrame` for consistency.

lib/session.js
```diff
@@ -413,7 +407,7 @@ Session.prototype._transferReceived = function(frame) {
   this._sessionParams.remoteOutgoingWindow--;
 
   if (this._sessionParams.incomingWindow < 0) {
-    // @todo Shut down session since sender is not respecting window.
+    // @todo: Shut down session since sender is not respecting window.
     debug('Transfer frame received when no incoming window remaining, should shut down session but for now being tolerant.');
```
noodlefrenzy: switch to console.warn

lib/types.js
```diff
@@ -82,7 +87,7 @@ function registerType(name, options) {
  */
 function listBuilder(list, bufb, codec, width) {
   if (!Array.isArray(list)) {
-    throw new errors.EncodingError('Unsure how to encode non-array as list');
+    throw new errors.EncodingError(list, 'Unsure how to encode non-array as list');
```
noodlefrenzy: Good change, I like including the value in the error, why didn't I think of that?

lib/types/composite_type.js
```diff
@@ -0,0 +1,129 @@
...
+    if (field.multiple === false) {
+      throw Error('field ' + field.name + ' does not support multiple values, got ' + JSON.stringify(value));
+    }
+
+    if (value.length === 0) return new ForcedType(field.type, null);
+    else if (value.length === 1) return new ForcedType(field.type, value[0]);
+    else {
+      var values = [], _len = value.length;
+      for (var i = 0; i < _len; ++i) {
+        values.push(new ForcedType(field.type, value[i]));
```
noodlefrenzy: should this be `values.push(this.wrapField(field, value[i]));`, in case we have e.g. multiple of type '*'?

lib/types/composite_type.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  // @todo: this is a cheesy way to determine if its a composite type already
+  if (value.toDescribedType && typeof value.toDescribedType === 'function') return value;
+
+  return new ForcedType(field.type, value);
+}
+
+function compositeAccessors(index, field) {
```
noodlefrenzy: This is some slick shit, I like it.

lib/frames.js
mbroadst: :heavy_check_mark: made them both `*` with a note to add `requires: deliveryType` in the future.

lib/sender_link.js
mbroadst: :heavy_check_mark:

lib/session.js
mbroadst: :heavy_check_mark:

lib/types/composite_type.js
mbroadst: yup, fixing.

lib/types/composite_type.js
mbroadst: can't take credit for it, this was all Gordon - prettttty cool idea

lib/types/composite_type.js
mbroadst: :heavy_check_mark:

lib/session.js
```diff
@@ -290,29 +285,28 @@ Session.prototype._nextHandle = function() {
...
-    default:
-      debug('not yet processing frames of type: ' + frame.constructor.name);
-  }
+  debug('processing frame: ', frame);
+  if (frame instanceof frames.EndFrame) return this._processEndFrame(frame);
+  else if (frame instanceof frames.AttachFrame) return this._processAttachFrame(frame);
+  else if (frame instanceof frames.DetachFrame) return this._processDetachFrame(frame);
+  else if (frame instanceof frames.FlowFrame) return this._processFlowFrame(frame);
+  else if (frame instanceof frames.TransferFrame) return this._processTransferFrame(frame);
+  else if (frame instanceof frames.DispositionFrame) return this._processDisposition(frame);
```
noodlefrenzy: `_processDispositionFrame`

mbroadst: whoops, fixed. :heavy_check_mark:

lib/types/message.js
```diff
@@ -1,347 +1,191 @@
...
+function defineCompositeSection(definition) {
+  var descriptor = { name: 'amqp:' + definition.name + ':list', code: definition.code };
+  var sectionDefinition = {
+    descriptor: definition.code,
+    encode: function(value, buffer) {
+      var _len = definition.fields.length, result = [];
+      for (var i = 0; i < _len; ++i) {
+        var field = definition.fields[i];
+        if (value && value.hasOwnProperty(field.name)) {
+          result[i] = field.type !== '*' ? new ForcedType(field.type, value[field.name]) : value[field.name];
```
noodlefrenzy: should these be a `wrapField` call in case value[field.name] is already the correct type?

mbroadst: yeah probably not a bad idea. As you can tell, there is a certain amount of duplication here for these types vs a regular composite type, but there were very subtle yet non-trivial difficulties and I was a little lazy. I can go ahead and export `wrapField` from `composite-type` and use that here

lib/types/message.js
```diff
@@ -1,347 +1,191 @@
...
+    { name: 'deliveryCount', type: 'uint', default: 0 }
+  ]
+});
+
+defineCompositeSection({
+  name: 'properties', code: 0x73,
+  fields: [
+    { name: 'messageId', type: '*' },
+    { name: 'userId', type: 'binary' },
+    { name: 'to', type: 'string' },
```
noodlefrenzy: technically, `to` and `replyTo` are type '*', but I've never seen anything other than string. Might want to just put a comment to that effect.

lib/types/message.js
```diff
@@ -1,347 +1,191 @@
...
-    codec.encode(new Data(this.body), buf);
-  } else if (Array.isArray(this.body)) {
-    codec.encode(new AMQPSequence(this.body), buf);
-  } else {
-    codec.encode(new AMQPValue(this.body), buf);
+  if (!!message.body) {
+    if (message.body instanceof Buffer) {
+      codec.encode(new DescribedType(0x75, message.body), buffer);  // Data
+    } else if (Array.isArray(message.body)) {
+      codec.encode(new DescribedType(0x76, message.body), buffer);  // AMQPSequence
```
noodlefrenzy: do we want to deal with arrays of buffers as multiple Data segments here? arrays of arrays as multiple sequences? We never did before, so I'm fine leaving it out, but with your refactor it makes it _so much easier_.

lib/types/terminus.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+var types = require('../types'),
+    Terminus = module.exports = {};
+
+Terminus.Source = types.defineComposite({
+  name: 'source', code: 0x28,
+  fields: [
+    { name: 'address', type: 'string' },
```
noodlefrenzy: address is another one where it's technically '*' but I've never seen anything but string. Comment to that effect?

lib/types/message.js
mbroadst: yeah so there are a number of these sorts of situations, where the type is `*` but thats just pointing to a `requires`, which in this case is `requires="address"`

`address` is defined as: `<type name="address-string" class="restricted" source="string" provides="address"/>`

I did provide a way to make aliases in the type system `types.js` but some of these I just got lazy on.

test/integration/servicebus/eventhubs/client.test.js
```diff
@@ -8,154 +8,153 @@ var AMQPClient = require('../../../../lib/index.js').Client,
...
+        'apache.org:selector-filter:string': translator(
+          ['described', ['symbol', 'apache.org:selector-filter:string'], ['string', 'amqp.annotation.x-opt-enqueuedtimeutc > ' + now]])
+      } } }
+    };
+
+    return test.client.connect(config.address)
+      .then(function() {
+        return createPartitionReceivers(test.client, config.partitionCount, config.receiverLinkPrefix, filterOptions);
+      })
+      .map(function(receiver) {
```
noodlefrenzy: much nicer, thx

test/unit/client.test.js
```diff
@@ -167,50 +177,54 @@ describe('Client', function() {
...
-      var frameOverhead = TransferFrame.FRAME_OVERHEAD + deliveryTag.length;
+      var frameOverhead = frames.TRANSFER_FRAME_OVERHEAD + deliveryTag.length;
       var idealMessageSize = testMaxFrameSize - frameOverhead;
       var messageCount = Math.ceil(messageBuffer.length / idealMessageSize);
-      var expected = [], idx = 0;
+      var expectedFrames = [], idx = 0;
       for (var i = 0; i < messageCount; ++i) {
-        expected.push(messageBuffer.slice(idx, idx + idealMessageSize));
+        var frame = new frames.TransferFrame({
+          channel: 1, handle: 0, deliveryId: 1, settled: false, deliveryTag: deliveryTag, more: true,
```
noodlefrenzy: shouldn't it be more: false on the last one?

test/unit/test_utilities.js
```diff
@@ -89,47 +82,52 @@ describe('Utilities', function() {
...
+    // it('should work for described types', function() {
+    //   var dt = new DescribedType(new AMQPSymbol('keyname'), 'value string');
+    //   var merged = u.deepMerge({ options: dt });
+    //   expect(merged.options).to.be.an.instanceOf(DescribedType);
+    //   expect(merged.options.descriptor).to.be.an.instanceOf(AMQPSymbol);
+    //   expect(merged.options.descriptor.contents).to.eql('keyname');
+    //   expect(merged.options.value).to.eql('value string');
+    // });
+
+    /*
```
noodlefrenzy: This is worrying. Should I be worried? The fact that the integration tests are still working makes me happy, but I'm still a bit worried.

test/unit/client.test.js
mbroadst: :heavy_check_mark:

package.json
```diff
@@ -8,10 +8,10 @@
   },
```
noodlefrenzy: Major version bump, plz.

test/unit/test_utilities.js
mbroadst: Just adding a quick note here that we discussed this:
  * AMQPSymbol is gone so these tests would need to be changed generally (not so important)
  * our `deepMerge` isn't really the best at making proper own-and-inherited copies and merges, in the meantime in places where this is actually used internally we now depend on lodash's `merge`, until we can refactor ours to do everything we need it to do

PR #211, opened 2016-02-20
refactor(long): directly embed encoding for ulong/long

mbroadst:
The first step towards dropping our dependency on node-int64. 

@noodlefrenzy while you're at it, you might want to take a look at the code I already submitted for ulong/long encoding. It should have gone into review, but I think I was a bit over-eager there (it's right next to the code in this PR, so it shouldn't be hard to seek out).

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -433,16 +433,17 @@ registerType('ulong', {
     encoding(0x44, constantValue(0)),
     encoding(0x80, {
       encoder: function(val, bufb) {
-        if (val instanceof Int64) {
-          bufb.appendBuffer(val.toBuffer(true));
-        } else {
-          if (val < 0xFFFFFFFF) {
-            bufb.appendUInt32BE(0x0);
-            bufb.appendUInt32BE(val);
-          } else {
-            bufb.appendBuffer((new Int64(val)).toBuffer());
-          }
+        if (val instanceof Int64) return bufb.appendBuffer(val.toBuffer(true));
+        if (val instanceof Buffer) return bufb.appendBuffer(val);
+        if (typeof val !== 'number' || !(val instanceof Number)) val = Number(val);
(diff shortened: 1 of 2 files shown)


lib/types.js
```diff
@@ -484,18 +485,30 @@ registerType('long', {
...
+        if (!Number.isFinite(val)) {
+          throw new errors.EncodingError(val, 'invalid number');
+        }
+
+        var abs = Math.abs(val);
+        var high = abs / MAX_UINT;
+        var low = abs % MAX_UINT;
+        if (val > 0) {
+          bufb.appendInt32BE(high);
+          bufb.appendInt32BE(low);
```
mbroadst: I think this should actually be `bufb.appendUInt32BE(low);` ?

lib/types.js
```diff
@@ -484,18 +485,30 @@ registerType('long', {
...
         } else {
-          throw new errors.EncodingError(val, 'Invalid encoding type for 64-bit value');
+          // need to write to a buffer in order to calculate the 2s complement
+          var data = new Buffer(8), carry = 1;
+          for (var i = 7; i >= 0; i--) {
+            var value = ((low & 0xff) ^ 0xff) + carry;
+            data[i] = value & 0xff;
+            low = (i === 4) ? high : low >>> 8;
```
noodlefrenzy: I don't think you should reuse low here, I'd rather see a new variable name (e.g. `var curbytes = low;` before the for loop).

lib/types.js
noodlefrenzy: I _think_ this is ok.

noodlefrenzy:
I would _love_ to see some tests here where you encode using this new logic and decode using Int64 (and vice-versa). That would guarantee you're at least as right as they are, and make sure we don't regress when I come in later and muck something up.

Issue #210, opened 2016-02-17
Add receive strategy with more collaborative flow-control


noodlefrenzy:
Add a receive strategy (possibly the default?) that only increases link credit by the amount of messages that have been ACK'd (i.e. have had dispo frames sent back to the sender).

PR #214, opened 2016-02-25
Receive flow control on ack

noodlefrenzy:
Fix for #210

### DIFF
--- a/lib/link.js
+++ b/lib/link.js
@@ -146,7 +146,7 @@ Link.prototype._attachReceived = function(attachFrame) {
   this.emit(Link.Attached, this);
   this._resolveAttachPromises(null, this);
 
-  this._checkCredit(/* initial = */ true);
+  this._checkCredit({ initial: true });
 };
 
 // default implementation does nothing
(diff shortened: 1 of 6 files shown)


lib/link.js
```diff
@@ -146,7 +146,7 @@ Link.prototype._attachReceived = function(attachFrame) {
   this.emit(Link.Attached, this);
   this._resolveAttachPromises(null, this);
 
-  this._checkCredit();
+  this._checkCredit(/* initial = */ true);
```
mbroadst: is there any detriment to doing: `this._checkCredit({ initial: true });` here? for the pure sake of readability.

noodlefrenzy: I don't think so other than object instantiation vs. boolean, but since it's only on attach it's not like this is inner-loop code.

mbroadst:
that's my only real comment, this is pretty straightforward. LGTM

mbroadst:
Other than it might be nice to have a test for it? That might be a bit involved though

Issue #216, opened 2016-02-26
Coercion of string to 'error' type failing


noodlefrenzy:
e.g. in `ReceiverLink.reject(message, error)`, the string is failing to coerce to an AMQP error with an exception thrown from codec.js:167.

PR #217, opened 2016-02-26
refactor(errors): support encoding of AMQPErrors

mbroadst:
Not too much craziness going on here, but definitely enough to
warrant comment:

  - error conditions have all been moved to the ErrorCondition enum
    module. They are enumerated both by legacy enum format
    (e.g. InternalError, or LinkRedirect), but also now by names
    used in the spec ('internal-error', 'link:redirect'). This
    opens us up to the possibility of users simply needing to use
    a string in order to send the required error.

  - errors can be specified by full use of the AMQPError type
    internally (it is not exported), as well as by a plain object
    (e.g. { condition: 'internal-error', description: 'test' }), or
    just a string (for the mandatory condition field).

  - any unknown errors will throw an EncodingError

  - tests for all of this are baked in

### DIFF
--- a/lib/types/composite_type.js
+++ b/lib/types/composite_type.js
@@ -23,11 +23,9 @@ function wrapField(field, value) {
     return null;
   }
 
-  if (field.type === '*') {
-    if (field.hasOwnProperty('requires'))
-       return (value instanceof field.requires) ? value : new field.requires(value);
-    return value;
-  }
+  if (field.hasOwnProperty('requires'))
+    return (value instanceof field.requires) ? value : new field.requires(value);
+  if (field.type === '*') return value;
 
   if (Array.isArray(value)) {
     if (field.multiple === false) {
--- a/test/unit/types/errors.test.js
+++ b/test/unit/types/errors.test.js
@@ -1,0 +1,88 @@
+'use strict';
+var AMQPError = require('../../../lib/types/amqp_error'),
+    ErrorCondition = require('../../../lib/types/error_condition'),
+    frames = require('../../../lib/frames'),
+    builder = require('buffer-builder'),
+    tu = require('../testing_utils'),
+    expect = require('chai').expect;
+
+describe('Types(Errors)', function() {
+  var expected = tu.buildBuffer([
+    0x00, 0x00, 0x00, 0x30,
+    0x02, 0x00, 0x00, 0x00,
+    0x00, 0x53, 0x18,
+      0xc0, 0x23, 0x01,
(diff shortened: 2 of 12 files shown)


mbroadst:
forgot to mention:

closes #216

mbroadst:
I should also mention that this includes a nice little example of using `requires` in the composite definition. This same tactic could be used everywhere else we have a "choice" field (e.g. `TransferFrame.rcvSettleMode`) for two main benefits:
  * ensure we aren't screwing up any types that are being sent
  * ensure incoming data is in a format we expect it to be
  * allow users to use a more idiomatic node style values (`auto`/`settle`, instead of our exported constants)

noodlefrenzy:
"allow users to use a more idiomatic node style values (auto/settle, instead of our exported constants)"
This is **awesome**

lib/types/composite_type.js
```diff
@@ -23,11 +23,9 @@ function wrapField(field, value) {
     return null;
   }
 
-  if (field.type === '*') {
-    if (field.hasOwnProperty('requires'))
-       return (value instanceof field.requires) ? value : new field.requires(value);
-    return value;
-  }
+  if (field.hasOwnProperty('requires'))
+    return (value instanceof field.requires) ? value : new field.requires(value);
```
noodlefrenzy: will this ever return true, since in this case requires is a function which returns ErrorCondition and not ErrorCondition itself?

mbroadst: this is for cases like `terminus.Source/Target`, those are in fact composite types and were the initial use case for `requires`

mbroadst: https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/frames.js#L172

noodlefrenzy: Yeah, was just realizing that now that I've made it down to the actual error condition definitions. I'm cool.

test/unit/types/errors.test.js
```diff
@@ -0,0 +1,88 @@
...
+      error: new AMQPError({ condition: 'internal-error', description: 'test' })
+    });
+
+    var actual = tu.convertFrameToBuffer(close);
+    expect(expected).to.eql(actual);
+  });
+
+  it('should encode using object, enum error condition', function() {
+    var close = new frames.CloseFrame({
+      error: { condition: 'internal-error', description: 'test' }
```
noodlefrenzy: condition should be enum

mbroadst: oops

PR #222, opened 2016-03-05
Split ServiceBus Queue tests by process version

noodlefrenzy:
… to avoid races/collisions in parallel travis runs.

Oh, and add a little extra debug information for some dispo frame (lack of) processing.

### DIFF
--- a/test/integration/servicebus/queues/client.test.js
+++ b/test/integration/servicebus/queues/client.test.js
@@ -19,8 +19,7 @@ describe('Queues', function() {
   });
 
   it('should connect, send, and receive a message', function(done) {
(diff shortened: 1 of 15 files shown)


test/integration/servicebus/queues/client.test.js
```diff
@@ -19,8 +19,7 @@ describe('Queues', function() {
   });
 
   it('should connect, send, and receive a message', function(done) {
-    expect(config.serviceBusHost, 'Required env vars not found in ' + Object.keys(process.env)).to.exist;
-
+    expect(config.address).to.exist;
```
mbroadst: seems to me that these could all just go into a `before` right before the whole test suite?

PR #223, opened 2016-03-05
policy refactor (round 1)

mbroadst:
Currently specifying a policy for the client is a little bit awkward if you don't want to use one of the pre-rolled policies. This involves using the exported `Policy.merge` method, and generally doesn't feel very much like idiomatic node.

This patch changes the behavior such that a few things now hold true:

  * All policies inherit from a base `Policy`, so we can easily do `instanceof` checks now
  * The client ctor now accepts a policy, as well as inline policy overrides. This removes the need to ever use `Policy.merge`.
  * Additionally, since the base `Policy` class accepts `overrides` as an optional argument we now open the future up for simply exporting the policy functions rather than instances 
  * Finally, it is assumed that the argument passed into the client's ctor is a policy or a policyOverride. If its just an override (not an `instanceof Policy`), then it is implied that you are just providing overrides and they will be `merge`d with the default policy

I call this "(round 1)" because of the 3rd point listed above, but also because we have an additional breaking change that would be more ideal: currently the `Policy` ctor returns an `Object.seal(this)`, which allows us to continue to use `Policy.merge`. If we instead exported the policy functions themselves, we could fully remove any need for `merge`, and properly lock down policies with `Object.freeze`. We determined this wasn't enough of a change to warrant a semver major bump, so we'll leave it as-is for now.


lib/policies/event_hub_policy.js
```diff
@@ -1,6 +1,2 @@
 'use strict';
-
-var u = require('../utilities'),
-    ServiceBusPolicy = require('./service_bus_policy');
-
-module.exports = u.deepMerge({}, ServiceBusPolicy);
+module.exports = require('./service_bus_policy');
```
noodlefrenzy: Should have done this a long time ago. I kept expecting EH to need some actual overrides :)

test/unit/client.test.js
```diff
@@ -17,9 +17,10 @@ var _ = require('lodash'),
 
     test = require('./test-fixture');
 
-DefaultPolicy.connect.options.containerId = 'test';
-DefaultPolicy.reconnect.retries = 0;
-DefaultPolicy.reconnect.forever = false;
+var TestPolicy = new Policy({
```
noodlefrenzy: Much nicer.

test/unit/client.test.js
```diff
@@ -285,9 +286,10 @@ describe('Client', function() {
       setTimeout(function() { return test.server.setup(); }, 10);
 
       var address = test.server.address();
-      test.client.policy.reconnect = {
-        retries: 5, strategy: 'fibonacci', forever: true
-      };
+      test.client.policy = test.client.policy = new Policy({
```
noodlefrenzy: `test.client.policy = test.client.policy = `

Man, that is _really_ well assigned.

noodlefrenzy:
Other than the double-assignment, LGTM. :shipit:

test/unit/client.test.js
mbroadst: lol.. right fixing

PR #224, opened 2016-03-05
fix(link-reattach): ensure sessions don't reattach after disconnect

### LINKED ISSUE #221: Client.disconnect leaves link re-attach timeouts running
If links set re-attach policies, client disconnect fails to clear those timeouts, leading to cases where the links attempt re-attach on a dead connection/session and causing:

https://github.com/Azure/azure-iot-sdks/issues/320

mbroadst:
* add a new REATTACHING state for Links
 * ensure that reattach attempts are cleared on disconnect

Attempts to fix #221

### DIFF
--- a/lib/link.js
+++ b/lib/link.js
@@ -26,9 +26,17 @@ function Link(session, handle, linkPolicy) {
     'DETACHED': {
       sendAttach: function() {
         return this.ATTACHING;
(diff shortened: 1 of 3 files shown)


noodlefrenzy:
LGTM :shipit:

PR #226, opened 2016-03-11
refactor(types): add restricted types as requires for composites

mbroadst:


lib/frames.js
```diff
@@ -207,7 +237,7 @@ frames.TransferFrame = defineFrame(FrameType.AMQP, {
     { name: 'messageFormat', type: 'message-format', default: 0 },
     { name: 'settled', type: 'boolean' },
     { name: 'more', type: 'boolean', default: false },
-    { name: 'rcvSettleMode', type: 'ubyte', default: 0 }, // @spec type: 'receiver-settle-mode` => `boolean` choice
+    { name: 'rcvSettleMode', type: 'ubyte', requires: receiverSettleMode, default: 0 },
```
noodlefrenzy: `default: 'auto'`

lib/frames.js
```diff
@@ -126,6 +126,38 @@ function defineFrame(type, definition) {
   return Frame;
 }
 
+// restricted type helpers
+function role(value) {
```
noodlefrenzy: Man, it really _feels_ like there should be a way to easily factor this into common code. I just can't see what it is (I keep seeing how I'd write it in C#).

mbroadst: yeah I mean, most of these are just "choices", so we could instead set `requires: { mapping: 'of', objects: 'here' }`, but I figured doing it with a function was very easy for now - and doesn't cost too much

lib/types/terminus.js
```diff
@@ -1,17 +1,40 @@
 'use strict';
 var types = require('../types'),
+    errors = require('../errors'),
     Terminus = module.exports = {};
 
+function terminusDurability(value) {
```
noodlefrenzy: If you want to be consistent, we should only accept ['none', 'configuration', 'unsettled-state'] and default to 'none' below.

mbroadst: well I've been trying to be consistent in terms of supporting the path where you might be using our constants still (like, the whole codebase), while also supporting string versions. But yeah I'll default it to `none` in the field definition

PR #235, opened 2016-04-28
Avoid jsonifying Buffers for ServiceBus and variants.

noodlefrenzy:
Attempted fix for https://github.com/Azure/azure-event-hubs/issues/118

### DIFF
--- a/lib/policies/service_bus_policy.js
+++ b/lib/policies/service_bus_policy.js
@@ -10,6 +10,7 @@ module.exports = new Policy({
       maxMessageSize: 10000, // Arbitrary choice
     },
     encoder: function(body) {
+      if (body instanceof Buffer) return body;
       var bodyStr = body;
       if (typeof body !== 'string') {
         bodyStr = JSON.stringify(body);


mbroadst:
LGTM

ThomasAnnerel:
Just tried thuis commit.
We het followig exception:
Unhandled rejection (<{"descriptor":{"buffer":{"type":"Buffe...>, no stack trace)

https://gist.github.com/ThomasAnnerel/548307a92c84e6888b777c131c593c2f

ThomasAnnerel:
I guys I found another problem in the servicebus policy
maxMessageSize: 10000, // Arbitrary choice

If you look at https://azure.microsoft.com/en-us/documentation/articles/service-bus-quotas/ we notice this should be 256 bytes

Kind regards
Thomas Annerel

mbroadst:
@ThomasAnnerel sorry you're still having trouble! I'll defer to @noodlefrenzy on the actual ServiceBus related stuff here, but it looks like  you're correct on that constant - would you be willing to put together a PR for that?

PR #244, opened 2016-07-06
change merge to deepCopy in Session.prototype.begin() to create a new…

pierreca:
… instance of the session policy object.

using `_.merge` only references the existing policy object, which means the global policy object gets modified instead of an instance specific to the session later in the code.

changing it to `u.deepCopy` makes sure the session gets its own instance.


mbroadst:
fixed in 2cb0797f58ddd0f6e1e6be6535a19aeae19def0b, thanks again for the report and fix!

PR #247, opened 2016-08-09
Make sure transport is destroyed only once if terminate is called multiple times

pierreca:
related to https://github.com/noodlefrenzy/node-amqp10-transport-ws/pull/5

Depending on transport implementation, it might be the case that `Connection.terminate()` is called multiple times. In the case of `node-amqp10-transport-ws`, the `end` event was being emitted  synchronously when calling `transport.end()`, which led to a second call to `terminate()`. While executing the second call the transport got destroyed. Then the first call resumes execution and was trying to call `destroy()` on the transport again leading to a `NullReferenceException`.

Using a temporary variable and a null check prevents this from happening and makes sure terminate() can be called multiple times but is effectively trying to destroy the transport only once.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -537,15 +537,18 @@ Connection.prototype._saslComplete = function(err) {
 };
 
 Connection.prototype._terminate = function(hasError) {
-  if (this._transport) {
+  var transport = this._transport;
+  this._transport = null;
+
+  if (transport) {
     this._sslOptions = null;
     this.sasl = null;
     this.address = null;
-    this._transport.end();
+    transport.end();
     if (hasError) {
-      this._transport.destroy();
+      transport.destroy();
     }
-    this._transport = null;
+    transport = null;
   }
   if (this._heartbeatInterval) {
     clearInterval(this._heartbeatInterval);


mbroadst:
@pierreca it seems to me that the best way to handle this would actually be to add a new state to the connection state machine so that duplicate events are "dropped" since they aren't valid paths from this new state (maybe `CLOSING` or something?). @noodlefrenzy what do you think?

noodlefrenzy:
@mbroadst I would argue that his changes make the code safer regardless of whether we add an additional state, so should be accepted. 

I like the idea of the new state, but I've been away from the code for long enough that I don't have a good sense for the impact of a new state on the system - I know we do a lot of `if state == X` checks.

mbroadst:
okay fair enough, I'll see if I can put together a unit test to trigger this later today so we're covered if/when we refactor this stuff.

mbroadst:
@pierreca I saw the other PR was merged for the transport, I was putting together a unit test for the main module here but do you think that solves your problem sufficiently already?

pierreca:
Yes, with that change published I don't miserably crash on ECONNRESET anymore ;) You're off the hook and can figure out a better solution than this PR if you want.

PR #251, opened 2016-09-27
Add support for SASL ANONYMOUS.

dnwe:
This is an initial proposal on how the client might expose a way of the user not supplying credentials, but desiring to use SASL ANONYMOUS rather than raw AMQP.

I couldn't immediately think of a good policy-based way of doing the opt-in for this, so I just went with allowing a Sasl object to be supplied in the connect policy. Feel free to suggest a better way of providing this.


mbroadst:
@dnwe would you mind rebasing this on master to get the tests passing

dnwe:
@mbroadst Done, and passing. :+1:

mbroadst:
@dnwe wouldn't it be enough to specify a `connect` policy option called e.g. `saslMechanisms` (or something far more descriptive) in order to communicate which mechanisms we wish to support from the user perspective? I think most of this PR is great, but I agree with your later sentiment in that we do go out of our way not to expose much of the internal code - I'd like to discuss ways of doing that here before just exposing `Sasl`

dnwe:
Discussed this with @PRESTONA the other week and he concurred, and suggested something along the following lines:

A ﻿⁠⁠⁠⁠`saslMechanisms`﻿⁠⁠⁠⁠ entry in the policy as you suggest where:
 - ﻿⁠⁠⁠⁠**null**﻿⁠⁠⁠⁠ means: ﻿⁠⁠infer from the other policy settings and value supplied to connect﻿⁠⁠ (behaviour exactly as we have today, the default in policy.js)
 - ﻿⁠⁠⁠⁠**NONE**﻿⁠⁠⁠⁠ forces you not to attempt any SASL, raw AMQP only - if username / password are in the URL, they are ignored.
 - ﻿⁠⁠⁠⁠**PLAIN﻿⁠⁠⁠⁠** forces you to try SASL plain - it's a client-side error not to have a username and password in the URL
 - ﻿⁠⁠⁠⁠**ANONYMOUS﻿⁠⁠⁠⁠** forces you to try SASL anonymous - if username / password are in the URL, they are ignored.
 - **ANYTHING_ELSE** throws unsupported mechanism error

We could optionally also log a warning about user+pass being ignored in those cases above?

dnwe:
@mbroadst I think this is ready for re-review now

lib/sasl.js
```diff
@@ -10,19 +10,34 @@ var debug = require('debug')('amqp10:sasl'),
...
+function Sasl(mechanism) {
+  if (mechanism && !u.includes(saslMechanisms, mechanism)) {
+    throw new errors.NotImplementedError(
+      'Only SASL PLAIN and ANONYMOUS are supported.');
+  }
+  this.mechanism = mechanism;
   this.receivedHeader = false;
 }
 
+Sasl.Mechanism = saslMechanisms;
```
mbroadst: should this be: `Sasl.Mechanisms = saslMechanisms` ? (note the added *s*)

mbroadst:
@dnwe sorry, was a bit side tracked there. Aside from one nitpick about wording in the comments, this looks great!

lib/sasl.js
dnwe: They definitely should be consistent, but actually I think they should probably both be singular rather than plural.

mbroadst: I don't follow, isn't this an array of mechanisms? (hence the plural)

dnwe: Hmm or maybe you are right and they should be plural... :thinking:

dnwe: Although really we're using it like an enum, so its the type of the value, so I think it should be singular.

dnwe:
@mbroadst no probs, commit added for review comment

lib/sasl.js
mbroadst: I read this line as incorrect:
```
if (!u.includes(Sasl.Mechanism, saslMechanism)) {
```
in that we should be checking if the `Sasl.Mechanisms` (array/plural) "includes" the given saslMechanism (singular instance).

But I'm really not going to be so nitpicky here, if you think it's right then let's do it

dnwe: Yeah so I think that was why I left that one plural originally, where we are using it as a list of possible items to find one item in. However, when we export it on the Sasl.Mechanism we are really just using it as a named set of constants that you can choose from. Hence you choose the PLAIN mechanism (singular) constant via Sasl.Mechanism.PLAIN

dnwe: Totally happy if you want to change it the other way before merging. I think either works.

mbroadst:
@dnwe hm, another failing in our CI. We aren't running those SB tests against CI runs so we didn't catch that these changes broke support over there somehow. Maybe @noodlefrenzy can help out here, or I can extract those values again from travis to run them locally.

mbroadst:
@dnwe @noodlefrenzy I figured the issue out, it was related to the refactor of using `connectPolicy` (introduced in this PR) instead of `this.policy.connect`. `AMQPClient._newConnection` now operates somewhat more "functionally" requiring that the connect policy is passed into the method, rather than depending on the local internal `policy` property.

dnwe:
@mbroadst nice find!

PR #252, opened 2016-09-27
whitespace-only fix indenting in lib/transport

dnwe:
Seemed to be using tabs rather than the 2 spaces that the rest of the
codebase uses.


mbroadst:
thanks!

PR #256, opened 2016-09-28
fix(detach): don't always reject promise if !close

dnwe:
Additional fixup for #254

### DIFF
--- a/lib/link.js
+++ b/lib/link.js
@@ -118,7 +118,11 @@ Link.prototype.detach = function(options) {
     self.once(Link.Detached, function(info) {
       self.removeListener(Link.ErrorReceived, onError);
       if (!!info.error) return reject(info.error);
-      if (!info.closed) return reject('link not closed');
+      if (!info.closed) {
+        if (!options || options.closed) {
+          return reject('link not closed');
+        }
+      }
       resolve();
     });
   });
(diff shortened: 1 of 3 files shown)


dnwe:
@mbroadst the `link.test.js` tests seem to be passing OK on Travis now with this PR

lib/link.js
```diff
@@ -118,7 +118,11 @@ Link.prototype.detach = function(options) {
     self.once(Link.Detached, function(info) {
       self.removeListener(Link.ErrorReceived, onError);
       if (!!info.error) return reject(info.error);
-      if (!info.closed) return reject('link not closed');
+      if (!info.closed) {
+        if (!options || options.closed) {
+          return reject('link not closed');
```
mbroadst: don't we actually want to `resolve` in this case, as the detach has successfully occurred even though the link was not closed?

dnwe: yeah that's what this is doing:

- if the 'Detached' event that was emitting indicates that the link was not closed (!info.closed)
- then if either no options {} was supplied to the .detach(..) call, OR the options had requested closed === true, then we reject the promise because we wanted a close and we didn't get one
- otherwise we fall through to the default resolve() behaviour

mbroadst: oh right of course, my bad

Issue #255, opened 2016-09-27
Unable to choose AMQP encoding for Buffer or Array types


dnwe:
Currently the `encodeMessage` function in `lib/types/message.js` makes the following encoding decisions based upon the type of the `message.body`
```js
if (message.body instanceof Buffer) {
  codec.encode(new DescribedType(0x75, message.body), buffer);  // Data
} else if (Array.isArray(message.body)) {
  codec.encode(new DescribedType(0x76, message.body), buffer);  // AMQPSequence
} else {
  codec.encode(new DescribedType(0x77, message.body), buffer);  // AMQPValue
}
```

However, it is equally valid to take Buffer data and put it into a Binary type within an AMQPValue, or to take Array data and put it into an Array/List in an AMQPValue. Whilst the `codec.encode(..)` supports both of these, there doesn't appear to be a policy-based way of opting into the behaviour of always stuffing things into an AMQPValue.

mbroadst:
related: https://github.com/noodlefrenzy/node-amqp10/issues/97

dnwe:
Fixed.

PR #257, opened 2016-09-28
feat(types): allow DescribedType to be passed as the message.body

dnwe:
Allow a user to create a specific DescribedType in their message.body and honour this at encoding time.

-- 

As a first pass, this exposes a way of satisfying the functionality desired in #255, although it does require use of the internal DescribedType object in the client application. 

An alternative solution would be to allow the user to specify in their policy an encoding map of JavaScript objects and their desired message formats, perhaps exposing constants for the 0x75, 0x76, 0x77 hex type values.

### DIFF
--- a/lib/types/message.js
+++ b/lib/types/message.js
@@ -175,7 +175,14 @@ module.exports.encodeMessage = function(message, buffer) {
     // @todo:
     //  Array[Buffer] => multiple Data segments
     //  Array[Array] => multiple AMQPSequence
-    if (message.body instanceof Buffer) {
+    if (message.body instanceof DescribedType) {
+      var descriptor = message.body.descriptor;
+      if (!u.includes([0x75, 0x76, 0x77], descriptor)) {
+        throw new errors.MalformedPayloadError(
+            'Invalid described type for message body: ', descriptor);
+      }
+      codec.encode(message.body, buffer);
+    } else if (message.body instanceof Buffer) {
       codec.encode(new DescribedType(0x75, message.body), buffer);  // Data
     } else if (Array.isArray(message.body)) {
       codec.encode(new DescribedType(0x76, message.body), buffer);  // AMQPSequence
--- a/lib/index.js
+++ b/lib/index.js
@@ -26,6 +26,7 @@ module.exports = {
   },
 
   TransportProvider: require('./transport'),
+  DescribedType: require('./types/described_type'),
 
   /**
    * translator, which allows you to translate from node-amqp-encoder'd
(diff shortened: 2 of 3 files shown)


mbroadst:
Yeah, hm I think this might be a little too generic for the requirements dictated by the spec?
```
The body consists of one of the following three choices: one or more data sections, one or more amqp-sequence sections, or a single amqp-value section.
```

Simply allowing any `DescribedType` would allow the user to potentially send a malformed message wouldn't it?  I think in order to satisfy your desire in #255 we need to be able to more accurately describe/wrap the raw data to give hints to the heuristic checks here. This might include exposing some data types such as `AMQPValue` or `AMQPSequence`, which I removed with the most recent types refactor. 

BTW, @noodlefrenzy did start an effort to ease complex type description over here: https://github.com/noodlefrenzy/node-amqp-encoder. We technically support that, but I would be amenable to a more comprehensive solution.

dnwe:
@mbroadst commit refreshed with a check of the descriptor as discussed, hopefully you're OK to merge this in

mbroadst:
@dnwe looks like you forgot to export `DescribedType` on this round

dnwe:
@mbroadst I don't think I actually ever did export it, you're suggesting exporting it from amqp_client.js ?

dnwe:
Added in 9e01851

mbroadst:
Hm, perhaps I misunderstood the intent then? Didn't you want to be able to have user facing control over what the body of the message was? I don't see how it's possible with this patch without somehow exposing that type, so if that was indeed the intent I though we were discussing exposing it [here](https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/index.js#L6)

lib/types/message.js
```diff
@@ -175,7 +175,14 @@ module.exports.encodeMessage = function(message, buffer) {
     // @todo:
     //  Array[Buffer] => multiple Data segments
     //  Array[Array] => multiple AMQPSequence
-    if (message.body instanceof Buffer) {
+    if (message.body instanceof DescribedType) {
+      var descriptor = message.body.descriptor;
+      if (!u.includes([0x75, 0x76, 0x77], descriptor)) {
+        throw new errors.MalformedPayloadError('Unknown DescribedType: ',
```
mbroadst: just a little nitpick, but I think a more informative error might suit the user here. Like "Invalid described type for body", or probably something much better worded 😄

dnwe:
@mbroadst fixed and squashed the commits as b470e2b

PR #260, opened 2016-10-11
refactor(debug): output all received frames

dnwe:
Previously we were logging a selection of received frames via the
_processFrame handler(s) in session.js, but this ignores (e.g.,) a
received OpenFrame (which is handled in connection.js). It seems more
consistent to be logging received frames within the framing debug where
we currently log all sent frames.

### DIFF
--- a/lib/frames.js
+++ b/lib/frames.js
@@ -115,6 +115,7 @@ frames.readFrame = function(buffer) {
     frame.channel = channel;
   }
 
+  debug('received frame: ', frame, ': ' + payloadBuffer.toString('hex'));
   return frame;
 };
 
--- a/lib/frames.js
+++ b/lib/frames.js
@@ -4,6 +4,7 @@ var Builder = require('buffer-builder'),
     errors = require('./errors'),
     codec = require('./codec'),
     debug = require('debug')('amqp10:framing'),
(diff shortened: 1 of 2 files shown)


mbroadst:
@dnwe good point. I've actually been meaning to wrangle our use of debug for a while now, I'm going to put together a quick PR and let's see if you agree with it.

dnwe:
@mbroadst incidentally, I quite like the [debug](https://github.com/visionmedia/debug) module, but it is a bit of a shame that it doesn't provide any customization over the [output formatting](https://github.com/visionmedia/debug/blob/39ecd87bcc145de5ca1cbea1bf4caed02c34d30a/node.js#L95-L96), e.g., to select a more conventional HH:mm:ss.SSS timestamp format

mbroadst:
@dnwe ah yes I didn't mean a replacement for `debug`, rather just our sort of messy "all over the place" current use of it. I'll put my ideas into this PR and we can discuss there!

mbroadst:
@dnwe actually do you have any interest in joining out gitter? We could have somewhat more in-depth discussion there if you have the time/interest

dnwe:
@mbroadst extra commit added as per discussion on gitter

PR #264, opened 2016-10-13
refactor(Connection): track `local` and `remote` performatives

mbroadst:
@dnwe brought up an interesting point in his previous PR, in that
we have a sort of "mixed" concept of parameters when it comes to
the initial connection handshake. This patch introduces the concept
of a `local` and `remote` storage for sent and received frames.


dnwe:
Though probably removed the commented out _params block before you do :)

mbroadst:
doh, thanks

PR #266, opened 2016-10-14
Changing github references to point to NodeMessengers instead of nood…

noodlefrenzy:
…lefrenzy.


noodlefrenzy:
Not sure about changing references to things like gitter and travis.

mbroadst:
@noodlefrenzy yeah travis will need to change, and I think its a pretty good idea for gitter to change as well. But those should happen basically all at the same time

pierreca:
closing for housekeeping. no move in order for now.

Issue #274, opened 2016-10-27
Error Decoding AMQP Array


princjef:
I am trying to receive a message whose body is an AMQP value with the following format:

```js
{ // map
    "expirations": [/*timestamp*/]
}
```

However, when the message comes in, I get the following exception while decoding (types.js, line 210):

```
AmqpMalformedPayloadError: malformed payload: Unable to decode value of 83 from buffer 0a01830000015806f2fbc7 at index 0 of array
```

Looking at [codec._readFullValue](https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/codec.js#L46), it looks like there is an off-by-one error when using `forcedCode` with fixed-width values, as is the case with arrays. The decoder tries to read length + 1 bytes, even though the code has already been passed in the buffer.

I have a change that should fix this. Will open a pull request

mbroadst:
fixed in #275

PR #275, opened 2016-10-27
Only decode extra byte if not forcedCode

princjef:
Addresses the bug in #274 by only reading an extra byte for fixed-width values if the code is not provided by `forcedCode`


mbroadst:
awesome, thanks for including a test. Is there a reason the test for binary data was removed?

princjef:
whoops. I moved around my test and must have accidentally deleted it. it's back now

mbroadst:
@princjef great, thanks!

Issue #277, opened 2016-10-27
Can't change source link address on `createSender()` or target link address on `createReceiver()`


princjef:
I have been trying to implement the request/response pattern in the [AMQP Management working draft](https://www.oasis-open.org/committees/download.php/52425/amqp-man-v1%200-wd08.pdf) on top of node-amqp10.

It works by using a sender/receiver link pair, where the sender sets the `replyTo` property to the address of the receiver so that the response is sent to the correct place. In the spec, the sender sets its `source` address to `null` and the `receiver` sets its `target` address to a unique value that can be used in the `replyTo` field on sent messages. However, in `createSender()` and `createReceiver()`, the source and target, respectively, are set to `'localhost'` and cannot be overridden.

I looked at `parseLinkAddress()` but it only applies to the opposite component of the address, so it doesn't seem like there is any way to override this. To get it working, I moved around the values in the policy merge to allow overriding. This is what it looks like for [`createReceiver()`](https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/amqp_client.js#L261):

``` js
  var linkName = u.linkName(address.name, policyOverrides),
      linkPolicy = u.deepMerge({
        attach: {
          name: linkName,
          source: { address: address.name }
        }
      }, policyOverrides, { attach: { target: { address: 'localhost' } } }, this.policy.receiverLink);
```

Any ideas/suggestions?

mbroadst:
@princjef have you checked out the source for [amqp10-rpc](https://github.com/mbroadst/amqp10-rpc) ? I think I'm doing something similar here: https://github.com/mbroadst/amqp10-rpc/blob/master/lib/rpc-client.js#L16. 

You should definitely be able to override any value using the `policyOverrides`, so if we have a problem there (perhaps `deepMerge`/`deepCopy` is striking again!) then we need to resolve that.

princjef:
@mbroadst I think I have the opposite. I need to specify the address for the link that I am creating (a.k.a sender -> source, receiver -> target). I believe dynamic tells the link on the opposite end (sender -> target, receiver -> source) to set its own address.

As for whether it is overridable, it looks like `deepMerge` gives precendence in the opposite order of `Object.assign`, overriding values specified in later arguments with those specified in earlier ones. The current linkPolicy merge for `createReceiver` looks like this:

```js
  var linkName = u.linkName(address.name, policyOverrides),
      linkPolicy = u.deepMerge({
        attach: {
          name: linkName,
[...]

PR #278, opened 2016-10-28
Allow link address to be overridden in createSender()/createReceiver()

princjef:
Addresses #277 by reordering the policy merge parameters on `createSender()` and `createReceiver()`. With this change, it is possible to override the source address to be overridden on `createSender()` and the target address on `createReceiver()`. This is relevant for use cases such as [AMQP request/response](https://www.oasis-open.org/committees/download.php/52425/amqp-man-v1%200-wd08.pdf?p=17).

The opposite address can already be overridden by other means, so it was not moved.


princjef:
superseded by #279

PR #290, opened 2017-01-27
Prevent exception if transport is already null when Connection.prototype._terminate() is called twice

pierreca:
Using node 0.10, when `Connection.prototype._terminate()` is called, `this._transport.end()` calls `socket.end()` which generates an event that leads to `_terminate()` being called again synchronously (meaning the first call to `_terminate()` is still waiting for `this._transport.end()` to return while the second call nullifies the `_transport` object. 

Once the second call is finished, the first call continues and throws an exception when trying to call `this._transport.terminate()`.

This PR should prevent that from happening :)

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/290)

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -530,15 +530,16 @@ Connection.prototype._saslComplete = function(err) {
 };
 
 Connection.prototype._terminate = function(hasError) {
-  if (this._transport) {
-    this._sslOptions = null;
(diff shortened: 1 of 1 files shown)


mbroadst:
@pierreca thanks for the fix, lgtm

PR #294, opened 2017-02-14
Add UUID support for encoder

pierreca:
related to https://github.com/noodlefrenzy/node-amqp-encoder/pull/2

The goal is to add the ability for the user to force the UUID type on properties, and especially on the messageId and correlationId properties.

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/294)

### DIFF
--- a/lib/adapters/translate_encoder.js
+++ b/lib/adapters/translate_encoder.js
@@ -30,6 +30,8 @@ var encoder = processor({
         return null;
       case 'boolean':
         return val;
+      case 'uuid':
+        return new ForcedType('uuid', val);
       default:
         throw new Error('Unknown fixed: ' + type);
     }


lib/adapters/translate_encoder.js
```diff
@@ -30,6 +30,8 @@ var encoder = processor({
         return null;
       case 'boolean':
         return val;
+      case 'uuid':
```
mbroadst: I think you can get away with just adding this to the first case above, the big one. Maybe we should just move that to the default case? it would have allowed you to do this without any changes if it was already that way

pierreca: Your call. it's a fixed width type according to the spec so I thought i belonged here but I'm fine adjusting that.

mbroadst: no you're right

mbroadst:
addressed in the merge of `encoder-poc` branch, which was just merged in

PR #308, opened 2017-03-24
Fix documentation to reference correct annotation field name

kurtb:
Some libraries that make use of this library (https://github.com/Azure/azure-event-hubs-node) were making use of the option field as "annotations" rather than "messageAnnotations." This caused failures when setting the partition key in Azure Event Hubs. Simple fix below to update some of the documentation. It looks correct in other spots. Not sure if the field changed at a certain point?

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/308)


mbroadst:
@kurtb ack, sorry about that. yeah definitely a legacy issue, corrected when we switched to using terms as close to spec as possible. Thanks!

PR #311, opened 2017-04-10
Add posibility to attach ReceiverLink delayed in code & createReceiverLink promise will resolve immediatelly

misak113:
* it is feature when you want to accept all messages even that it already waiting in queue. Otherwise the immediatelly attach will invoke receiving messages yarlier then event emmitter on('messasge'... can be registered.
* it has result that first messages are overlooked & never be consumed

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/311)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+<a name="3.5.1"></a>
+## [3.5.1](https://github.com/noodlefrenzy/node-amqp10/compare/v3.5.0...v3.5.1) (2017-03-14)
+
+
+
 <a name="3.5.0"></a>
 # [3.5.0](https://github.com/noodlefrenzy/node-amqp10/compare/v3.4.2...v3.5.0) (2017-03-07)
 
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -313,7 +313,11 @@ AMQPClient.prototype.createReceiver = function(address, policyOverrides) {
       };
 
       var link = self._session.createLink(linkPolicy);
-      link._onAttach.push(attachPromise);
+      if (!linkPolicy.attach.manually) {
+        link._onAttach.push(attachPromise);
+      } else {
+        attachPromise(null, link);
+      }
     };
 
     attach();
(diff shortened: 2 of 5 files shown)


mbroadst:
@misak113 can you explain in more detail what this is attempting to accomplish? The reason the attachPromise is pushed to `_onAttach` is that we don't want to resolve the promise until the link has actually been attached (in order to ensure order)

misak113:
I found the problem, when starts consuming from queue (RabbitMQ) where already exists some messages waiting for consumption.
Immediately after createReceiver (internally) all messages are consumed however Promise with created receiverLink is resolved (returned) after it. So it is not possible to subscribe EventEmitter for "message" before all messages in queue are consumed.
I'm using RabbitMQ with enabled plugin for AMQP 1.0 protocol version.

I try to show it in code:
```ts
export async function bindOne<TPayload extends IEventPayload>(
	client: Client, eventType: string,
	onEvent: (event: IEvent<TPayload>) => Promise<void>
) {
	const queueName = QUEUE_NAME_PREFIX + eventType;
// THIS part is problematic, because after resolving this line are messages already consumed (emitted to receiverLink EventEmitter)
	const receiver = await client.createReceiver(queueName);
// So this line would consume only new messages appeared in queue. All consumed messages are kept as not acked messages in queue & will be released after receiver is destroyed.
	receiver.on('message', async (message: Message) => {
		try {
			const event = message.body;
			await onEvent(event);
			receiver.accept(message);
		} catch (error) {
			receiver.reject(message);
			throw error;
		}
	});
	receiver.on('errorReceived', (error: Error) => console.error(error));
// this line is necessary to be added if I apply this fixing commit & use receiver policy with manually attach
	receiver.attach();
}
```

vIceBerg:
Hi!

What about this change? I face the same issue and would like it to be resolved....

amarzavery:
@vIceBerg - This repository is not being actively maintained. I suggest using [rhea](https://github.com/amqp/rhea).

vIceBerg:
Hi!

I know... I'm waiting for your Pull Request on Azure SDK for Service Bus using AMQP... Meanwhile, I try to find some solutions to temporarily patch my problems... This was one of them.

I ended up forking the repo and copy/pasted the changes.

PR #317, opened 2017-05-08
fix(recv-link): distinguish auto/manual settlement

### LINKED ISSUE #316: rcvSettleMode is mapping `first` => `autoSettle` and `second` => `manually settle`
Currently the amqp10 module seems to have re-purposed AMQP's rcv-settle-mode on a receiver link to determine whether or not the client library itself should be doing automatic settlement when messages received, or whether to expect the user to call `.accept()` to settle received messages.

However, I think this is incorrect. The rcv-settle-mode FIRST vs SECOND is specifically governing the contract between sender and receiver link as to who MUST send the settlement disposition first.

From the spec

> rcv-settle-mode
If first, this indicates that the Receiver MUST settle the delivery once it has arrived
without waiting for the Sender to settle first.
If second, this indicates that the Receiver MUST NOT settle until sending its disposition
to the Sender and receiving a settled disposition from the sender.
If not set, this value is defaulted to the value negotiated on link attach.
If the negotiated link value is first, then it is illegal to set this field to second.
If the message is being sent settled by the Sender, the value of this field is ignored.
The (implicit or explicit) value of this field does not form part of the transfer state,
and is not retained if a link is suspended and subsequently resumed.

Currently the 'SECOND' behaviour isn't implemented by the module, as per:
```
ReceiverLink.prototype._dispositionReceived = function(details) {
  debug('not yet processing "receiver" disposition frames: ', details);
};
```

I think for now the right thing would be to permanently lock rcv-settle-mode to FIRST to have AMQP.Constants.receiverSettleMode.autoSettle and AMQP.Constants.receiverSettleMode.manualSettle to simply distinguish whether or not the user plans to settle messages themselves.

dnwe:
This should probably be a major version bump as the API is changed

- default to rcv-settle-mode "first" everywhere, we don't currently handle
  "second" correctly anyway
- add link policy attribute "settlement" to distinguish between auto,
  where the client will immediately accept messages on receipt, and manual
  where the user is expected to call accept when they have finished
  handling a delivery.

Fixes #316

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/317)

### DIFF
--- a/lib/receiver_link.js
+++ b/lib/receiver_link.js
@@ -192,7 +192,7 @@ ReceiverLink.prototype._messageReceived = function(transferFrame) {
   // @todo: Bump link credit based on strategy
 
   // respect settle mode in policy
-  if (this.policy.settlement === constants.settlement.manual) {
+  if (this.policy.settlement === constants.settlement.auto) {
     this.accept(message);
   }
 
--- a/lib/receiver_link.js
+++ b/lib/receiver_link.js
@@ -20,7 +20,7 @@ var util = require('util'),
 function ReceiverLink(session, handle, linkPolicy) {
   ReceiverLink.super_.call(this, session, handle, linkPolicy);
   // If we're not auto-settling, we should track settles to allow refresh policies to decide when to re-up.
-  this._trackSettles = linkPolicy && linkPolicy.attach && linkPolicy.attach.rcvSettleMode !== constants.receiverSettleMode.autoSettle;
+  this._trackSettles = linkPolicy && linkPolicy.settlement === constants.settlement.manual;
   this._currentTransferFrame = null;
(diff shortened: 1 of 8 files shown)


lib/receiver_link.js
```diff
@@ -192,7 +192,7 @@ ReceiverLink.prototype._messageReceived = function(transferFrame) {
   // @todo: Bump link credit based on strategy
 
   // respect settle mode in policy
-  if (this.policy.attach.rcvSettleMode === constants.receiverSettleMode.autoSettle) {
+  if (this.policy.settlement === constants.settlement.manual) {
```
mmaestri: actually this should be
`if (this.policy.settlement === constants.settlement.manual) {`

PR #319, opened 2017-05-09
fix(address-parsing): don't truncate ':' passwords

### LINKED ISSUE #318: Default policy doesn't copy with username/password's that contain a colon (:)
The default policy truncates passwords at the first colon (':') character that they contain. Obviously this means that you can't use passwords that contain this character to successfully connect to a remote server.

This is due to the way that `lib/policies/policy.js` parses a the value from a `urlObject.auth` field.  Specifically:
```javascript
    var userPass = parsedAddress.auth.split(':', 2);
    result.user = userPass[0];
    result.pass = userPass[1] || null;
 ```

Because `urlObject.auth` is always a URI decoded value - if either the username or password portion of the authentication information contain a colon character it is ambiguous as to where the two fields are delimited.

dnwe:
Fix a bug which truncated passwords containing a ':' character because
the parseAddress function relied upon node's url.parse().auth which
returns the user:pass pair already URI-decoded.

Fixes #318

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/319)

### DIFF
--- a/lib/policies/policy.js
+++ b/lib/policies/policy.js
@@ -237,11 +237,13 @@ Policy.prototype.parseAddress = function(address) {
 
   result.rootUri = result.protocol + '://';
   if (!!parsedAddress.auth) {
-    result.rootUri += parsedAddress.auth + '@';
-
-    var userPass = parsedAddress.auth.split(':', 2);
-    result.user = userPass[0];
-    result.pass = userPass[1] || null;
+    // capture the part of the encoded URI between '//' and '@'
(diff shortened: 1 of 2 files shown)


mbroadst:
@dnwe thanks for the contribution, sorry I've been out of town for the past few weeks.

mbroadst:
@prestona @dnwe looks like this breaks on earlier versions of node, any chance you can update to accommodate that?

Issue #322, opened 2017-05-17
TypeError: Cannot read property 'write' of null


jtjk:
Wed May 17 2017 19:19:23 GMT+0000 (Coordinated Universal Time): Application has thrown an uncaught exception and is terminated:
TypeError: Cannot read property 'write' of null
    at Object.frames.writeFrame (D:\home\site\wwwroot\node_modules\amqp10\lib\frames.js:56:9)
    at Connection.sendFrame (D:\home\site\wwwroot\node_modules\amqp10\lib\connection.js:328:10)
    at ReceiverLink.Link.attach (D:\home\site\wwwroot\node_modules\amqp10\lib\link.js:137:27)
    at ReceiverLink.Link._attemptReattach (D:\home\site\wwwroot\node_modules\amqp10\lib\link.js:252:8)
    at Timeout._onTimeout (D:\home\site\wwwroot\node_modules\amqp10\lib\link.js:240:10)
    at ontimeout (timers.js:365:14)
    at tryOnTimeout (timers.js:237:5)
    at Timer.listOnTimeout (timers.js:207:5)

I guess this is well known issue. This should be fixed. I am trying to connect to Azure IoT Hub

jtjk:
I founded workaround:
frames.js lines 52->
  var buffer = builder.get();
  buffer.writeUInt32BE(buffer.length, 0);
  debug('sending frame: ', frame);
  trace('raw: [' + buffer.toString('hex') + ']');
  **if (stream != null)**
    stream.write(buffer, callback);

Yes, this workaround is insane, but I don't have time to debug what is the root cause for that so I use this workaround until something better is coming :)

jtjk:
[https://github.com/noodlefrenzy/node-amqp10/pull/323](url)

jtjk:
Any comments? I am still getting this occasionally

sebastus:
I also get this error quite frequently. Great to see the workaround. Would really prefer an actual fix.

PR #323, opened 2017-05-17
Only write stream if stream exists

jtjk:
[https://github.com/noodlefrenzy/node-amqp10/issues/322](url)

Not so elegant workaround to fix linked issue

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/323)

### DIFF
--- a/lib/frames.js
+++ b/lib/frames.js
@@ -60,8 +60,9 @@ frames.writeFrame = function(frame, stream, options, callback) {
     debug('=>', frame);
     trace('raw: [' + buffer.toString('hex') + ']');
   }
-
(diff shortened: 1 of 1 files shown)


jtjk:
Any comments to this?

pierreca:
@jtjk I think the problem is in the state machine rather than in the call to stream.write(), this would just swallow an error that at this point is perfectly legitimate (albeit annoying).
Not denying there might be a bug, just saying this probably isn't the right way to fix it, therefore i'm closing the PR but feel free to reopen it if you want to discuss it further.

PR #329, opened 2017-07-30
update tests to support different runtime versions

bengreenier:
- [x] fix breaking tests
- [ ] update semver patch before shipping
- [x] remove `.travis.yml` changes

A [couple of tests](https://travis-ci.org/noodlefrenzy/node-amqp10/builds/234319960) fail on the master branch, as a result of downlevel changes in node core. Specifically, `node>=5.7.0` adds [encodeAuth()](https://github.com/nodejs/node/commit/db9128135f382e22202d900263e872d8b48f0050#diff-bbc5176adff1bbc01acd61bfc85d049fR924) (see [changelog](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V5.md#2016-02-23-version-570-stable-rvagg) for which release it landed in - `5.7.0`) which inteligently __doesn't__ encode `:` as a uri component when creating a [Url](https://nodejs.org/api/url.html#url_class_url). Rather than make any changes to our library to normalize this behavior across versions (which feels like a bad fix, since this behavior is consistent with core) we added support conditionally executing tests based on a version check against the current runtime.

We apply this new version checking behavior to the failing test, and provide an additional new test - the result is a version for `node < 5.7.0` and a __different__ version for `node >= 5.7.0`, to test both expected behaviors on the respective runtimes.

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/329)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,8 @@
   "devDependencies": {
     "benchmark": "^2.1.2",
     "chai": "^3.5.0",
-    "chai-string": "^1.3.0",
     "chai-as-promised": "^6.0.0",
+    "chai-string": "^1.3.0",
     "conventional-changelog-cli": "^1.2.0",
     "gh-pages": "^0.12.0",
     "istanbul": "^0.4.5",
@@ -27,6 +27,7 @@
     "jshint": "^2.9.4",
     "mocha": "^3.2.0",
     "qmf2": "^0.1.5",
+    "semver": "^5.4.1",
     "stream-buffers": "^3.0.1"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "amqp10",
-  "version": "3.5.3",
+  "version": "3.5.4",
   "description": "Native AMQP-1.0 client for node.js",
   "main": "./lib",
   "engines": {
--- a/test/testing_utils.js
+++ b/test/testing_utils.js
@@ -4,6 +4,7 @@ var Builder = require('buffer-builder'),
     expect = require('chai').expect,
     _ = require('lodash'),
     sb = require('stream-buffers'),
+    semver = require('semver'),
     frames = require('../lib/frames');
 
 function populateConfig(configKeyMap, cb) {
@@ -120,7 +121,6 @@ module.exports.convertFrameToBuffer = function(frame) {
   return buffer.getContents();
 };
 
-
 module.exports.bufferEqual = function(a, b) {
   if (!Buffer.isBuffer(a)) return undefined;
   if (!Buffer.isBuffer(b)) return undefined;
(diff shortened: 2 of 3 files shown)


bengreenier:
hmm, this is triggering:

```
Creating home directory `/var/run/qpid' ...
/etc/init.d/qpidd: 52: /etc/init.d/qpidd: source: not found
user had insufficient privilege
invoke-rc.d: initscript qpidd, action "start" failed.
dpkg: error processing package qpidd (--configure):
```

I'll continue to look into it - i suspect something about how the qpid deb installs itself has changed since the last official build...

mbroadst:
@bengreenier thanks for the contribution. I don't agree with the changes to `.travis.yml` (not sure why they are necessary), and I think perhaps the solution is a little too heavy handed. afaict, the change introduced in #319 is what broke this, and it was only broken for ancient versions of node (0.10/0.12), which are only presently still supported because of some existing MS customers.

It should be enough to simply modify [this code](https://github.com/noodlefrenzy/node-amqp10/pull/319/files#diff-5ae435bcf166b7f6645a061f699d242dL240) to accommodate both cases (maybe do a string check against `process.version` or something). Your solution would simply not run the test against older versions of node, and thus continue to (silently) break on those versions.

bengreenier:
@mbroadst anytime - been talking to @noodlefrenzy about getting more involved here. 😄 

The `.travis.yml` changes will go away, was just trying to figure out why the build doesn't work. I abstracted the `process.version` string check into a test helper because i thought this might come up again and that solution would be easier to re-purpose. I'm not just doing a string compare because it seems more brittle than pulling in `semver` (since it's just a devDep, i figured there's little impact).

Would you prefer less complexity (like an inline `process.version` check) as opposed to this approach?

FWIW, my current approach isn't just a skip for that test, it modifies the existing to only run on `>=5.7.0` and adds a new one that runs on `<5.7.0` - since they're effectively testing different behaviors.

mbroadst:
@bengreenier oh I see, sorry I missed the second test. I'm still surprised the existing code in `policy.js` isn't broken for older versions of node? Otherwise, yeah, LGTM once the `travis.yml` is reverted

bengreenier:
@mbroadst seems like the `policy.js` code works because the issue only impacts the `url.href` field, not `url.auth` - interestingly enough the change was in how formatting of `url.auth` is done into the `url.href` field, as i understand it.

sounds good, i'll revert that stuff - any idea why the ci build fails? i was seeing the failure with and without my travis changes.

mbroadst:
@bengreenier I'd have to see the logs for when you revert travis, right now it looks like permissions error with installing qpid - but that shouldn't be the case with the old config.

bengreenier:
@mbroadst sure, i'll get the reverted by EOD. 

Though in the meantime, the first PR build (`b899f2b`) had no `.travis.yml` changes and still had that issue - https://travis-ci.org/noodlefrenzy/node-amqp10/builds/259141707 

though If you'd prefer to wait til i revert and see if it repros that makes sense as well.

bengreenier:
removed those travis changes entirely, lets see if the build clears.

mbroadst:
@bengreenier looks like maybe the qpid maintainers screwed up the 1.36 release - unfortunately it will take me a bit to get to that. In the meantime you could maybe shoot an email over to iboverma@redhat.com indicating there's a problem with their ubuntu packaging (perhaps try to install it locally on an ubuntu server running trusty first?). Sorry for the trouble!

bengreenier:
@mbroadst no worries, I'll investigate and ping them - will update this pr as needed.

Issue #331, opened 2017-09-06
Cannot create multiple sessions per connection


princjef:
As far as I can tell, creating a client using the current amqp10 implementation creates [a single session](https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/amqp_client.js#L70), which becomes tied to the client/connection and is used for all links created on that connection.

While this works well for most cases, I chatted with the Azure Service Bus about some problems that we've been seeing related to https://github.com/noodlefrenzy/node-amqp10/issues/245 and it turns out that Service Bus expects almost every link to exist within its own session (though there are some specialized cases where two links should share a session). This has significant implications around connection/disconnection of links (and general communication - Service Bus doesn't design/test for single session per connection so there are no guarantees that it won't break in the future). It turns out that for many cases where Service Bus sends a detach on a link (such as when a client is idle for too long), they invalidate the *entire session*, rather than just the *individual link*. As a result, reconnecting the link doesn't always work as expected and other links that still seem to be attached can start misbehaving or simply stop responding. While all of this is going on, the actual connection is still sending heartbeats back and forth just fine.

I'm going to devote a lot more time to this tomorrow to try to think through a reasonable design to this which can support this need without making the usage pattern for other clients more complicated. My initial thought is to add something like `createSession()` to the `AMQPClient` class, which would create a new session and then expose `createSender()` and `createReceiver()` off of whatever object gets returned. It wouldn't be as simple as that (that wouldn't handle reattach for the link vs. the session, for instance), but it seems like a reasonable starting point. Anyway, I'd be happy to hear any thoughts/concerns about this and will update with more once I've had more of a chance to think through the implications of such a change.

(*Side note: the [AMQP Protocol Guide](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-protocol-guide) for Service Bus states that there should only be one session per connection, but I have confirmation from the Service Bus team that this is not the case and that the documentation is simply out of date/wrong*)

princjef:
Alright I've poked around the code some more and have a slightly better idea as to what this would entail. Here's my current thinking:

 1. **Add a `createSession()` method to the `AMQPClient` class.** I'm not passing any parameters for now, but it could potentially take a `policyOverrides` for the `session` portion of the client's policy. Just like when creating links, the client must first be connected before a session can be created. The method returns a promise which resolves with a `Session` object once the session is mapped.
[...]

PR #330, opened 2017-08-16
Update .travis.yml

oferb1:
Since v1.36.0 for qpidd, the package fails to build, due to issues with the `init.d` (see https://issues.apache.org/jira/browse/QPID-7863)
As a result travis has been failing
As a work around, I have added a basic `upstart` script, and moved the installation of `qpidd` and `qpid-tools` to the `before_install` section of the travis file

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/330)

PR #332, opened 2017-09-07
feat(session): add support for multiple sessions per client

princjef:
Addresses https://github.com/noodlefrenzy/node-amqp10/issues/331 by adding the ability to create multiple sessions on the same client/connection and providing policy options for reestablishing the session on end. This feature set is designed to better align the capabilities of amqp10 with the session pattern supported by Azure Service Bus and bring the retry logic inline with what is used for reattach and reconnect.

**Features**
 - [x] Add `createSession()` in `AMQPClient` to create user sessions
    - [x] Support custom policy overrides
 - [x] Allow an optional session to be passed in on `createSender()`, `createSenderStream()`, `createReceiver()`, `createReceiverStream()`
 - [x] Create the default session associated with the client only when needed
 - [x] `restart` policy option for reestablishing the session after it is ended

**Tests**
 - [x] `createSession()`
 - [x] Optional session parameter for `createSender()`, `createSenderStream()`, `createReceiver()`, `createReceiverStream()`
 - [x] Creating default client session on demand
 - [x] Session policy - `restart`

Please take a look and let me know what you think of the change. I'm open to changing the design/approach if people have other thoughts.

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/332)

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -68,6 +68,8 @@ function AMQPClient(policy, policyOverrides) {
 
   this._connection = null;
   this._session = null;
+  this._sessionPromise = null;
+  this._userSessions = [];
 
   this._reconnect = null;
   if (!!this.policy.reconnect) {
@@ -157,20 +159,54 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
     self._connection.on(Connection.Connected, function(c) {
       debug('connected');
       self.emit(AMQPClient.ConnectionOpened);
+
+      var promises = [];
+      
+      // Only map the session if it has already been created
       if (self._session) {
         debug('session already exists, re-using');
         self._session.connection = self._connection;
-      } else {
-        self._session = self._newSession(c);
+
+        promises.push(new Promise(function(resolve, reject) {
+          self._session.once(Session.Mapped, function(s) {
+            debug('client session mapped');
+            resolve();
+          });
+
+          self._session.begin(pu.Merge(self.policy.session, self._session.policy));
+        }));
       }
 
-      self._session.once(Session.Mapped, function(s) {
-        debug('mapped');
+      // Update any user-generated sessions
+      promises = promises.concat(self._userSessions.map(function(session) {
+        session.connection = self._connection;
+
+        return new Promise(function(resolve, reject) {
+          session.once(Session.Mapped, function(s) {
+            debug('user session mapped');
+            resolve();
+          });
+          
+          // We specifically don't merge the policies here because the policy
+          // for these sessions is passed in explicitly on create of the
+          // session. It's very unlikely that there wouldn't be a policy for the
+          // session at this point, but we fall back to the session policy from
+          // the client just in case. That being said, the policy specified
+          // explicitly by the user for that session should take precendence,
+          // leading to the lack of a merge here.
+          session.begin(session.policy || self.policy.session);
+        });
+      }));
+
+      if (promises.length > 0) {
+        Promise.all(promises).then(function() {
+          self.emit('connected');
+          resolve(self); 
+        }).catch(reject);
+      } else {
         self.emit('connected');
         resolve(self);
-      });
-
-      self._session.begin(self.policy.session);
+      }
     });
 
     self._connection.on(Connection.Disconnected, function() {
@@ -195,6 +231,35 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
 };
 
 /**
+ * Creates a session for the current connection that can be associated with any
+ * new links on that connection
+ *
+ * @inner @memberof AMQPClient
+ *
+ * @return {Promise<Session>}
+ */
+AMQPClient.prototype.createSession = function(policyOverrides) {
+  if (!this._connection) {
+    throw new Error('Must connect before creating sessions');
+  }
+
+  policyOverrides = policyOverrides || {};
+
+  var session = this._newSession(this._connection);
+  this._userSessions.push(session);
+
+  var self = this;
+  return new Promise(function (resolve, reject) {
+    session.once(Session.Mapped, function() {
+      debug('user session mapped');
+      resolve(session);
+    });
+
+    session.begin(pu.Merge(policyOverrides, self.policy.session));
+  });
+};
+
+/**
  * Creates a sender link for the given address, with optional link policy
  *
  * @inner @memberof AMQPClient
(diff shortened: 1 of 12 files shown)


princjef:
I also pulled in the change from https://github.com/noodlefrenzy/node-amqp10/pull/330 so that I can get useful information from the Travis CI build. That change looks good to me and I think it's worth merging in before this goes in.

princjef:
Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[lib/policies/policy.js, line 127 at r1](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/332#-KtYL6jCZoBOuylDEv59:-KtYL6jDJvs9qf_x_03n:b7o2v0p) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/94f9f91ef92500fa3091b01dfb0614d0103d07f4/lib/policies/policy.js#L127)):*
> ```JavaScript
>       enableSessionFlowControl: true,
>       closeOnLinkError: false,
>       reattach: null
> ```

I'm not wild about calling this `reattach`, since the concept of attach/detach applies to links, not sessions. The equivalent for the session terminology would be `rebegin` or `remap`, but neither really feels right to me. Thoughts?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/332)*

princjef:
I decided to remove the `endOnLinkDetach` policy from the session because it's not a part of the spec and was causing more problems than it was solving. The other changes have stood up well to my testing though

lib/amqp_client.js
```diff
@@ -157,20 +159,47 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
...
-      } else {
-        self._session = self._newSession(c);
+
+        promises.push(new Promise(function(resolve, reject) {
+          self._session.once(Session.Mapped, function(s) {
+            debug('client session mapped');
+            resolve();
+          });
+
+          self._session.begin(self._session.policy || self.policy.session);
```
pierreca: >self._session.policy || self.policy.session [](start = 30, length = 43)

if these 2 are the same, no problem. if the existing session doesn't have a policy object, apply the configured one. i get that part - but what if self._session.policy is set to something that is incompatible with self.policy.session (which i assume is the desirable policy) ? should these objects be merged instead ? or why keep the existing policy object at all?

lib/amqp_client.js
```diff
@@ -157,20 +159,47 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
...
+            resolve();
+          });
+          
+          session.begin(session.policy || self.policy.session);
+        });
+      }));
+
+      if (promises.length > 0) {
+        Promise.all(promises).then(function() {
+          self.emit('connected');
```
pierreca: >self.emit('connected'); [](start = 10, length = 23)

unrelated but puzzling nonetheless to me: why would the "connected" state of the client be attached to sessions rather than connection or links?

princjef: yeah this one is an artifact of how the Promise-based design works. it's trying to make sure things are in a usable state when the promise returns. since the session is baked into the connection by the library, that means both the connection and the session have to be ready. i considered making it lazy, but didn't know that i'd be able to get away with it without some subtle breaking changes

princjef: actually i did make it lazy. the problem is reconnect because the session is already there

test/unit/test_session.js
```diff
@@ -22,7 +22,7 @@ var test = {
     connect: { options: { containerId: 'test' } },
     senderLink: { attach: { name: 'sender' } },
     receiverLink: { attach: { name: 'receiver' } }
-  }, DefaultPolicy)
+  }, DefaultPolicy),
```
pierreca: >, [](start = 19, length = 1)

typo?

pierreca:
reestablish? the spec talks about establishing and ending sessions

---
In reply to: [328216141](https://github.com/noodlefrenzy/node-amqp10/pull/332#issuecomment-328216141) [](ancestors = 328216141)

lib/amqp_client.js
pierreca: ha I see. the philosophy of policies as i understand it though is that they are merged and what is passed around are policy overrides - so starting with the client policy and then merging/overriding with the session policy would feel like the natural thing to do. but i defer to your judgement.

---
In reply to: [163131314](https://github.com/noodlefrenzy/node-amqp10/pull/332#discussion_r163131314) [](ancestors = 163131314)

test/unit/test_session.js
princjef: probably had something else and took it out. good catch

lib/amqp_client.js
princjef: i think it's reasonable to do the merge on the built-in one (this line). the user one is weird because the `begin()` here is a hidden operation. the policy for the session is specified separately so i don't want to override it if it's already there. i'll add a comment to that one

princjef:
ah i like that. reestablish sounds better to me than restart

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/332#-:-L3Z8rIzOshwXZSssS6W:b5t7yw)*

princjef:
*[lib/amqp_client.js, line 176 at r2](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/332#-KtOpGL0EBIOYMuErCZR-r2-176:-L3ZAHvSeaOt8jqoL9MD:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/75e080f56030dcb6deb2fe47ec93fe013023f44c/lib/amqp_client.js#L176)):*
<details><summary><i>Previously, princjef (Jeff Principe) wrote…</i></summary><blockquote>

i think it's reasonable to do the merge on the built-in one (this line). the user one is weird because the `begin()` here is a hidden operation. the policy for the session is specified separately so i don't want to override it if it's already there. i'll add a comment to that one
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/332)*

princjef:
*[test/unit/test_session.js, line 25 at r2](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/332#-KtywJnpgl_IqeHmjqrE-r2-25:-L3ZAXeeFJBkLurfz0WR:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/75e080f56030dcb6deb2fe47ec93fe013023f44c/test/unit/test_session.js#L25)):*
<details><summary><i>Previously, princjef (Jeff Principe) wrote…</i></summary><blockquote>

probably had something else and took it out. good catch
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/332)*

PR #345, opened 2018-01-08
Update amqp_client.js with guard around reconnect

siraajahmed:
Stops the client attempting to reconnect when the client has been disconnected on purpose.

When disconnect has been called, `_preventReconnect()` gets called which sets `_reconnect = null`.
This guard has been added to stop the client from attempting to reconnect which will end with an error from `_attemptReconnection()` with regards to `return self._reconnect()` --  self._reconnect is not a function -- because it has been set to null

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/345)

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -380,7 +380,7 @@ AMQPClient.prototype.disconnect = function() {
       });
       connection.close();
     } else {
-      debug("already disconnected")
+      debug("already disconnected");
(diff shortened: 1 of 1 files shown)


pierreca:
@siraajahmed I'd like to see a test to validate this case - is it something you could add to this PR?

siraajahmed:
Not going to have time

PR #349, opened 2018-01-25
Separate SASL mechanism and add "registerSaslMechanism" API on the client

pierreca:
This adds the ability to "inject" a SASL mechanism and associated handler in the library, which effectively unlocks the ability to have complex SASL challenge/response exchanges.

- moved sasl-anonymous and and sasl-plain logic within their own handlers (classes)
- modified client & connection to enable injecting a handler in the main sasl logic
- added a client API (`registerSaslMechanism`) to enable the user to inject their own sasl handler.

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -5,7 +5,9 @@ var EventEmitter = require('events').EventEmitter,
     util = require('util'),
 
     Connection = require('./connection'),
-    Sasl = require('./sasl'),
+    Sasl = require('./sasl/sasl'),
+    SaslPlain = require('./sasl/sasl_plain'),
+    SaslAnonymous = require('./sasl/sasl_anonymous'),
     Session = require('./session'),
     ReceiverStream = require('./streams/receiver_stream'),
     SenderStream = require('./streams/sender_stream'),
@@ -65,7 +67,7 @@ function AMQPClient(policy, policyOverrides) {
     this.policy = policy || new Policy();
     if (!!policyOverrides) this.policy = pu.Merge(policyOverrides, this.policy);
   }
-
+  this.saslHandlers = {};
   this._connection = null;
   this._session = null;
   this._sessionPromise = null;
@@ -130,7 +132,7 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
     var saslMechanism = connectPolicy.saslMechanism;
     var sasl = null;
     if (saslMechanism) {
-      if (!u.includes(Sasl.Mechanism, saslMechanism)) {
+      if (!u.includes(Sasl.Mechanism, saslMechanism) && !self.saslHandlers[saslMechanism]) {
         throw new errors.NotImplementedError(
           saslMechanism + ' is not a supported saslMechanism policy');
       }
@@ -143,12 +145,21 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
         throw new errors.AuthenticationError(
             'Sasl PLAIN requested, but no credentials provided in endpoint URI');
       } else {
-        sasl = new Sasl(saslMechanism);
+        if (self.saslHandlers[saslMechanism]) {
+          sasl = new Sasl(saslMechanism, self.saslHandlers[saslMechanism]);
+        } else if (saslMechanism === Sasl.Mechanism.ANONYMOUS) {
+            sasl = new Sasl(Sasl.Mechanism.ANONYMOUS, new SaslAnonymous());
+        } else if (saslMechanism === Sasl.Mechanism.PLAIN) {
+          sasl = new Sasl(Sasl.Mechanism.PLAIN, new SaslPlain());
+        } else {
+          throw new errors.AuthenticationError(
+            'No adequate SASL handler for the requested mechanism');
+        }
       }
     } else if (address.user) {
       // force SASL plain if no mechanism specified, but creds in URI
       connectPolicy.saslMechanism = Sasl.Mechanism.PLAIN;
-      sasl = new Sasl(Sasl.Mechanism.PLAIN);
+      sasl = new Sasl(Sasl.Mechanism.PLAIN, new SaslPlain());
     }
     if (!!sasl && !!address.vhost) {
       sasl._remoteHostname = address.vhost;
@@ -161,7 +172,7 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
       self.emit(AMQPClient.ConnectionOpened);
 
       var promises = [];
-      
+
       // Only map the session if it has already been created
       if (self._session) {
         debug('session already exists, re-using');
@@ -186,7 +197,7 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
             debug('user session mapped');
             resolve();
           });
-          
+
           // We specifically don't merge the policies here because the policy
           // for these sessions is passed in explicitly on create of the
           // session. It's very unlikely that there wouldn't be a policy for the
@@ -201,7 +212,7 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
       if (promises.length > 0) {
         Promise.all(promises).then(function() {
           self.emit('connected');
-          resolve(self); 
+          resolve(self);
         }).catch(reject);
       } else {
         self.emit('connected');
@@ -434,6 +445,17 @@ AMQPClient.prototype.disconnect = function() {
       resolve(); // Already disconnected, just deliver the promise.
     }
   });
+};
+
+/**
+ * Registers a new SASL mechanism to handle SASL challenges during authentication.
+ *
+ * @param {string}      mechanism   Name of the mechanism used to authenticate - must match what the server sends in the SASL-Mechanism frame.
+ * @param {SaslHandler} handler     Object that is going to handle SASL challenges and craft corresponding responses.
+ */
+AMQPClient.prototype.registerSaslMechanism = function(mechanism, handler) {
+  Sasl.Mechanism[mechanism] = mechanism;
+  this.saslHandlers[mechanism] = handler;
 };
 
 // Use the provided session, falling back to the client's session and creating
--- a/lib/sasl/sasl.js
+++ b/lib/sasl/sasl.js
@@ -1,14 +1,11 @@
 'use strict';
 
 var debug = require('debug')('amqp10:sasl'),
-    Builder = require('buffer-builder'),
-
-    constants = require('./constants'),
-    frames = require('./frames'),
-    errors = require('./errors'),
-    u = require('./utilities'),
-
-    Connection = require('./connection');
+    constants = require('../constants'),
+    frames = require('../frames'),
+    errors = require('../errors'),
+    u = require('../utilities'),
+    Connection = require('../connection');
 
 var saslMechanism = {
   PLAIN: 'PLAIN',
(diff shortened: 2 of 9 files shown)


lib/sasl/sasl.js
```diff
@@ -21,12 +18,12 @@ var saslMechanism = {
  *
  * @constructor
  */
-function Sasl(mechanism) {
-  if (mechanism && !u.includes(saslMechanism, mechanism)) {
-    throw new errors.NotImplementedError(
-      'Only SASL PLAIN and ANONYMOUS are supported.');
+function Sasl(mechanism, handler) {
```
amarzavery: - Can you please document the shape of the handler? 
- Should we verify for certain property names in the handler and verify whether they are of type `function`?
- What happens if someone provides an empty object `{}` for the handler? I assume it will fail later.
- Clarifying and documenting above things will help me provide better typings for sasl handler

amarzavery:
Travis is failing for some weird reason in 0.10 version of node.js. Have restarted the ci job. Btw, do we want to support 0.10 or should we remove this from the build matrix?

amarzavery:
Otherwise everything LGTM

princjef:
travis is failing for 0.10 because there are a couple of arrow functions

amarzavery:
oops. @pierreca can you replace the arrow functions?

lib/sasl/sasl.js
```diff
@@ -60,41 +54,50 @@ Sasl.prototype.headerReceived = function(header) {
...
-            'Sasl ANONYMOUS requested, but credentials provided in endpoint URI');
-      }
-      buf.appendUInt8(0); // <null>
-    } else {
-      this.callback(new errors.NotImplementedError(
-          'Only SASL PLAIN and ANONYMOUS are supported.'));
+
+    if (mechanism === Sasl.Mechanism.ANONYMOUS && this.credentials.user && this.credentials.pass) {
+      console.warn(
+          'Sasl ANONYMOUS requested, but credentials provided in endpoint URI');
```
princjef: does it make sense for this to be in the sasl anonymous handler instead so that the knowledge of that mechanism is self-contained?

lib/sasl/sasl_anonymous.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+var Builder = require('buffer-builder'),
+    Promise = require('bluebird');
+
+function SaslAnonymous () {}
+
+SaslAnonymous.prototype.getInitFrameContent = function() {
+  return new Promise(function(resolve) {
```
princjef: nit: in my experience it's cleaner doing `Promise.resolve(value)` rather than newing a Promise if you don't need to wrap any asynchronous operations

lib/amqp_client.js
```diff
@@ -65,7 +67,7 @@ function AMQPClient(policy, policyOverrides) {
     this.policy = policy || new Policy();
     if (!!policyOverrides) this.policy = pu.Merge(policyOverrides, this.policy);
   }
-
+  this.saslHandlers = {};
```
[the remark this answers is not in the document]
pierreca: absolutely, just sloppiness on my part here.

lib/sasl/sasl.js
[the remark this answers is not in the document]
pierreca: it's not exported so probably not a breaking change.

pierreca: I'm going to add property/function checks in the public API, and i'm adding samples and docs too.

lib/sasl/sasl.js
pierreca: I've gone back and forth on this. Ideally, yes, but since the credentials object is not a thing that is part of the Sasl-Anonymous mechanism it would also be weird for it to be there IMHO. Plus, it feels safer to not move that for now since it was already a part of the existing logic (making sure we're not breaking too many things at once).

lib/sasl/sasl.js
```diff
@@ -60,41 +54,50 @@ Sasl.prototype.headerReceived = function(header) {
...
-    var initFrame = new frames.SaslInitFrame({
-      mechanism: mechanism,
-      initialResponse: buf.get()
-    });
 
-    if (!!this._remoteHostname) initFrame.hostname = this._remoteHostname;
-    this.connection.sendFrame(initFrame);
+    this.handler.getInitFrameContent(self.credentials)
+      .then(function (initFrameContent) {
+        var initFrame = new frames.SaslInitFrame({
```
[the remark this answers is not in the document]
pierreca: that's exactly why. I'm not sure we want to start exporting our frames types. It's worth revisiting this decision now though since changing it later would likely be a breaking change.

lib/sasl/sasl_anonymous.js
pierreca: Fair enough. i'm not very familiar with Promises (we don't use them) so i'm glad to get this type of feedback. thanks!

lib/sasl/sasl.js
princjef: i think that's reasonable

lib/sasl/sasl_plain.js
```diff
@@ -0,0 +1,29 @@
...
+
+    resolve({
+      mechanism: 'PLAIN',
+      initialResponse: buf.get()
+    });
+  });
+};
+
+SaslPlain.prototype.getChallengeResponseContent = function () {
+  return new Promise(function(resolve) {
```
amarzavery: This is much simpler and does the same thing
```javascript
return Promise.resolve();
```

lib/sasl/sasl_anonymous.js
```diff
@@ -0,0 +1,24 @@
...
+    buf.appendUInt8(0); // <null>
+    resolve({
+      mechanism: 'ANONYMOUS',
+      initialResponse: buf.get()
+    });
+  });
+};
+
+SaslAnonymous.prototype.getChallengeResponseContent = function () {
+  return new Promise(function(resolve) {
```
amarzavery: same as below

pierreca:
Review status: 0 of 11 files reviewed at latest revision, 6 unresolved discussions.

---

*[lib/sasl/sasl.js, line 74 at r1](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349#-L3fFA-OjzQXThsc9cw5-r1-74:-L3kRtoYeJyQeMRaB8PU:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/6d0edc556aa31997f0a7ad70fce065b506edd26e/lib/sasl/sasl.js#L74)):*
<details><summary><i>Previously, pierreca (Pierre Cauchois) wrote…</i></summary><blockquote>

that's exactly why. I'm not sure we want to start exporting our frames types. It's worth revisiting this decision now though since changing it later would likely be a breaking change.
</blockquote></details>

Done.

---

*[lib/sasl/sasl.js, line 91 at r1](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349#-L3fFA-OjzQXThsc9cw5-r1-91:-L3kRvJ-azpZ6lWHt737:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/6d0edc556aa31997f0a7ad70fce065b506edd26e/lib/sasl/sasl.js#L91)):*
<details><summary><i>Previously, princjef (Jeff Principe) wrote…</i></summary><blockquote>

any reason for using an arrow function + `self`? i'm guessing this is just a typo and you meant to do `function (responseContent) {` instead (though i'd be all for switching to arrow functions if/when we drop support for old versions of node).

same for the `.catch` after this
</blockquote></details>

Done.

---

*[lib/sasl/sasl.js, line 69 at r2](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349#-L3fFA-OjzQXThsc9cw5-r2-69:-L3kRw7RSaII06zU7IVU:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/1cd0a93de3bc16e7a247d159b8aa96403534e07c/lib/sasl/sasl.js#L69)):*
<details><summary><i>Previously, princjef (Jeff Principe) wrote…</i></summary><blockquote>

i think that's reasonable
</blockquote></details>

Done.

---

*[lib/sasl/sasl_anonymous.js, line 19 at r2](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349#-L3fFA-OjzQXThsc9cw6-r2-19:-L3kS00xWcP83Rg5IJM_:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/1cd0a93de3bc16e7a247d159b8aa96403534e07c/lib/sasl/sasl_anonymous.js#L19)):*
<details><summary><i>Previously, amarzavery (Amar Zavery) wrote…</i></summary><blockquote>

same as below
</blockquote></details>

Done.

---

*[lib/sasl/sasl_plain.js, line 24 at r2](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349#-L3fFA-OjzQXThsc9cw7-r2-24:-L3kS2jzEwgy8NacTvam:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/1cd0a93de3bc16e7a247d159b8aa96403534e07c/lib/sasl/sasl_plain.js#L24)):*
<details><summary><i>Previously, amarzavery (Amar Zavery) wrote…</i></summary><blockquote>

This is much simpler and does the same thing
```javascript
return Promise.resolve();
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349)*

lib/sasl/sasl.js
```diff
@@ -60,41 +54,53 @@ Sasl.prototype.headerReceived = function(header) {
...
+
+        self.connection.sendFrame(initFrame);
+      })
+      .catch(function (err) {
+        self.callback(new errors.AuthenticationError('SASL Init Failed: ' + frame.code + ': ' + frame.details + ' with error: ' + err.toString()));
+      });
   } else if (frame instanceof frames.SaslChallengeFrame) {
-    var responseFrame = new frames.SaslResponseFrame({});
-    this.connection.sendFrame(responseFrame);
+    if (typeof (this.handler.getResponseFrame) === 'function') {
```
princjef: nit: parentheses shouldn't be necessary

princjef:
Reviewed 2 of 10 files at r1, 9 of 9 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349)*

princjef:
LGTM

pierreca:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[lib/sasl/sasl.js, line 90 at r4](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349#-L3fFA-OjzQXThsc9cw5-r4-90:-L3ni4bVj9ZoeW5okmBW:b-896fix) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/82afa479c5a1e7bf08699aa9bb325fe022548e1e/lib/sasl/sasl.js#L90)):*
<details><summary><i>Previously, princjef (Jeff Principe) wrote…</i></summary><blockquote>

nit: parentheses shouldn't be necessary
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/349)*

PR #352, opened 2018-01-30
fix(session): clear session states between reconnections

pierreca:
Here's a (blunt) suggestion on how to clean sessions between reconnections - I'm not sure in which circumstances it's desirable to keep a session over multiple reconnections.

Without this some tests start failing with IoT Hub when trying to reattach links over "persisted" sessions.

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/352)

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -203,7 +203,7 @@ Session.prototype.createLink = function(linkPolicy) {
   var self = this;
   link.on(Link.Detached, function(details) {
     debug('detached(' + link.name + '): ' + (details ? details.error : 'No details'));
-    if (!link.shouldReattach() || self.disposed) self._removeLink(link);
+    if (!link.shouldReattach() || self._disposed) self._removeLink(link);
   });
 
   link.on(Link.ErrorReceived, function(err) {
@@ -477,9 +477,9 @@ Session.prototype._processDispositionFrame = function(frame) {
  */
 Session.prototype._resetLinkState = function() {
   var self = this;
-  var forceDetachLink = function(l) { 
+  var forceDetachLink = function(l) {
     l.forceDetach();
-    if (self.disposed) self._removeLink(l);
+    if (self._disposed) self._removeLink(l);
   };
   u.values(this._senderLinks).forEach(forceDetachLink);
   u.values(this._receiverLinks).forEach(forceDetachLink);
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -518,10 +518,17 @@ AMQPClient.prototype._clearConnectionState = function(saveReconnectDetails) {
     this._reconnect = null;
   }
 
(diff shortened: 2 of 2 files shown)


lib/session.js
```diff
@@ -203,7 +203,7 @@ Session.prototype.createLink = function(linkPolicy) {
   var self = this;
   link.on(Link.Detached, function(details) {
     debug('detached(' + link.name + '): ' + (details ? details.error : 'No details'));
-    if (!link.shouldReattach() || self.disposed) self._removeLink(link);
+    if (!link.shouldReattach() || self._disposed) self._removeLink(link);
```
amarzavery: Were we calling the wrong property on the object before?

amarzavery:
CI is failing
```
 1) Client #reconnect should wait for all existing sessions to be mapped before completing:


      Uncaught AssertionError: expected 0 to equal 1

      + expected - actual


      -0

      +1

      

      at assertEqual (node_modules/chai/lib/chai/core/assertions.js:487:12)

      at ctx.(anonymous function) (node_modules/chai/lib/chai/utils/addMethod.js:41:25)

      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:293:29)

      at null.<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:252:17)

      at ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/overwriteMethod.js:49:33)

      at AMQPClient.<anonymous> (test/unit/client.test.js:762:45)

      at Connection.<anonymous> (lib/amqp_client.js:9:9213)

      at Connection._processOpenFrame (lib/connection.js:9:19902)

      at Connection._receiveAny (lib/connection.js:9:15737)

      at Connection._receiveData (lib/connection.js:9:11006)

      at null.<anonymous> (lib/connection.js:9:22386)

      at Socket.<anonymous> (lib/transport/net_transport.js:9:1586)

      at Socket.<anonymous> (_stream_readable.js:765:14)

      at emitReadable_ (_stream_readable.js:427:10)

      at emitReadable (_stream_readable.js:423:5)

      at readableAddChunk (_stream_readable.js:166:9)

      at Socket.Readable.push (_stream_readable.js:128:10)

      at TCP.onread (net.js:529:21)
```

lib/session.js
princjef: yeah that's my bad...

princjef: actually wait the naming is fine. i defined a getter on the class called `disposed` which is the value of `this._disposed` so this should be fine

princjef: see line 156

pierreca: agree - this is more of a nitpicking thing / just for consistency purposes.

pierreca:
wrong fix - new PR coming up from @princjef

PR #353, opened 2018-01-30
fix(session): disable reestablish on disconnect and guard against duplicate begin() calls

princjef:
With `reestablish` enabled for sessions, the session can end up in a state where it is trying to reestablish itself while the connection is trying to do the same, leading to duplicate calls to `begin()`. The following has been changed to better handle this situation and others like it:

 * Disable the `begin()` attempt (and any further retries) if the link isn't still `UNMAPPED`. We already handle if the connection is broken, but this handles the connection coming down and back up (with a corresponding `begin()` call) while we're waiting to retry.
 * Guard against initializing a session multiple times. Since `begin()` isn't idempotent, calling it on a session that is already active with attached links will cause a second session to be established and the links to be implicitly assigned to that session without the server being informed. We now do nothing if the session is already active and throw if it's in a transition state.
 * Unmap a session if its connection is disconnected. This was already being done implicitly, but the state machine wasn't actually updating to the new state, which was erroneously tripping the state check in `begin()`. The `disconnected()` transition has been added to the state machine to take care of this. For reference, the AMQP 1.0 spec states that "Sessions end automatically when the Connection is closed or interrupted.", so this behavior is consistent with the specification.

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/353)

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -18,16 +18,50 @@ var EventEmitter = require('events').EventEmitter,
     Connection = require('./connection');
 
 var stateMachine = function(session) {
+  function warnBeginState(state) {
+    console.warn('BUG: Tried to call begin() on a session in nonterminal state ' + state);
+  }
+
   return {
-    'UNMAPPED': { sendBegin: 'BEGIN_SENT' },
-    'BEGIN_SENT': { beginReceived: 'MAPPED' },
+    'UNMAPPED': {
+      begin: function(sessionPolicy) {
+        session._performBegin(sessionPolicy);
+        return this.BEGIN_SENT;
+      }
+    },
+    'BEGIN_SENT': {
+      beginReceived: 'MAPPED',
+      disconnected: 'UNMAPPED',
+      begin: function(sessionPolicy) {
+        warnBeginState('BEGIN_SENT');
+      }
+    },
     'MAPPED': {
       sendEnd: 'END_SENT',
-      endReceived: 'END_RCVD'
+      endReceived: 'END_RCVD',
+      disconnected: 'UNMAPPED'
     },
-    'END_SENT': { endReceived: 'UNMAPPED' },
-    'DISCARDING': { endReceived: 'UNMAPPED' },
-    'END_RCVD': { sendEnd: 'UNMAPPED' }
+    'END_SENT': {
+      endReceived: 'UNMAPPED',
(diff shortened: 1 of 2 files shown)


pierreca:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[lib/session.js, line 181 at r1](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/353#-L48UanhR5-M93IqGyWi:-L48UanhR5-M93IqGyWj:b-pv0rag) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/52ff5e398c6fe092dcd48378e9d0e91ef429f7f7/lib/session.js#L181)):*
> ```JavaScript
>   var state = this.sm.getMachineState();
>   debug('trying to begin() a session in state ' + state);
>   switch (state) {
> ```

wouldn't that switch mean that we need to make "begin" an action of the state machine?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/353)*

princjef:
*[lib/session.js, line 181 at r1](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/353#-L48UanhR5-M93IqGyWi:-L48ljzR-wgcsuvq1zOQ:b66bevp) ([raw file](https://github.com/noodlefrenzy/node-amqp10/blob/52ff5e398c6fe092dcd48378e9d0e91ef429f7f7/lib/session.js#L181)):*
<details><summary><i>Previously, pierreca (Pierre Cauchois) wrote…</i></summary><blockquote>

wouldn't that switch mean that we need to make "begin" an action of the state machine?
</blockquote></details>

good call. moved `begin()` into the state machine in place of `beginSent()` with the appropriate guards/checks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/353)*

pierreca:
:lgtm:

---

Reviewed 1 of 1 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/noodlefrenzy/node-amqp10/353#-:-L49-28vFSHvI-EeMc0R:bnfp4nl)*

PR #354, opened 2018-01-31
fix(session): unmap session when connection is updated

princjef:
When the connection is changed in a session, it is effectively unmapped since the connection for which it is mapped is no longer associated with it. This wasn't an issue (besides the accumulation of AMQP sessions) until we started validating the session state on `begin()`. This change watches for changes to the connection for the session and updates the state accordingly to make the behavior consistent with the previous behavior.

As a side note: I tried to use the existing test harness to write a unit test exercising this case, but could not get it to behave and tear down correctly after several hours so I'm giving up on it for now. I have verified the fix manually for now.

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/354)


amarzavery:
Lgtm

PR #364, opened 2018-03-12
fixes #358 - the connection:opened event should be emited after all operations ?

### LINKED ISSUE #358: Duplicate on message handlers.
I have to use logic in a 'connection:opened' handler to send a sas token. 

If I do anything with the client in this 'connection:opened' event handler, I get two calls to my "on message" handler, regardless of where my receiver was created. I have verified I am only creating one receiver, once.

karlvlam:
Creating receiver link right after 'connection:opened' event will get duplicated messages. Using process.nextTick() to create the link can avoid the issue. 

Please review whether this fix is appropriate or not...

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/364)

### DIFF
--- a/lib/amqp_client.js
+++ b/lib/amqp_client.js
@@ -169,7 +169,6 @@ AMQPClient.prototype.connect = function(url, policyOverrides) {
     self._connection = self._newConnection(connectPolicy);
     self._connection.on(Connection.Connected, function(c) {
       debug('connected');
-      self.emit(AMQPClient.ConnectionOpened);
 
       var promises = [];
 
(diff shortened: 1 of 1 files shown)


tony-gutierrez:
@noodlefrenzy Can we get some integrations of these fixes?!

amarzavery (approved):

amarzavery:
@princjef - What do you think?

princjef:
The current behavior should not be a new thing. Though I don't know the original intent of the code, my understanding of the `connection:opened` event is that it specifically indicates that the AMQP connection was opened successfully. To that end, the event is currently being fired in the right place.

That being said, you typically need at least a connection+session to really do anything in AMQP. That's what the `connected` event is for. It tells you when your client is ready to work with links, etc. There isn't a constant for it in the client, but you should be able to listen to that instead of modifying the semantics of the `connection:opened` event. Give that a shot and let me know if it resolves your issue.

For reference, the `connected` event is fired at https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/amqp_client.js#L214 and https://github.com/noodlefrenzy/node-amqp10/blob/master/lib/amqp_client.js#L218.

PR #367, opened 2018-06-08
Update policy.js

hodonsky:
Trying to connect to AmazonMQ requirement of the connection protocol to be `amqp+ssl`

---
This change is [Reviewable](https://reviewable.io/reviews/noodlefrenzy/node-amqp10/367)


hodonsky:
Ah shoot this is incomplete.... I will PR again when I add it to the accepted protocols and things

